PR #6, opened 2013-07-30
Fixed problem in 0.10 where onListen gets called twice

mattinsler:


hbetts:
@mattinsler has this issue been resolved in the latest `0.10.x` release (which is 0.10.40)?

mattinsler:
Haven't tried it in a while. Started to use other ways of picking a port.

PR #7, opened 2014-03-01
Add a node bin script for calling this logic from shell

awinder:
I'd like to use this code in a CI environment to grab an unused port on the machine so that I can spin up the same codebase on multiple ports (CI concurrency FTW!). Rather than just write this logic into my codebase I figured this could be useful for others.


hbetts:
:+1:

PR #12, opened 2015-02-06
require async to use it

cades:
as title


indexzero:
Was cherry-picked. Thanks! Went into `0.4.0`

PR #14, opened 2015-05-23
Set license to "MIT" in package.json

kemitchell:
This tiny pull request replaces `MIT/X11` with just `MIT` in package.json, so you won't get a validation warnings about a non-[SPDX](http://spdx.org/licenses) license identifier.


kemitchell:
Thanks!

PR #20, opened 2016-01-23
Don't allow yielding of a port that is in use on a different interface

### LINKED ISSUE #13: portfinder yields a port that is already in use
So, this issue might be related to #9. I'm seeing this on Mac OS X:
```bash
➜  ~/.local/android-sdk-macosx/platform-tools/adb forward --list
3739ced5 tcp:8000 localfilesystem:/data/local/debugger-socket
➜  curl http://127.0.0.1:8000
(expected output from the android device)
➜  node -e 'require("portfinder").getPort(console.log)'
null 8000
```
In other words, I have port 8000 forwarded to an Android device via `adb` and there's a service running that responds to curl. But, portfinder does not detect it, and it offers 8000 as a free port. 

This doesn't seem limited to `adb`. Per #9, I tried running a PHP dev server and got similar results:
```bash
➜ php -S 127.0.0.1:8000
PHP 5.5.14 Development Server started at Thu Mar 19 15:41:33 2015
Listening on http://127.0.0.1:8000
Document root is /Users/lmorchard/devel/mozilla/node-firefox/node-firefox-forward-ports
Press Ctrl-C to quit.
```
(switch to another shell)
```bash
➜ curl -si http://127.0.0.1:8000 | head -1
HTTP/1.1 404 Not Found
➜ node -e 'require("portfinder").getPort(console.log)'
null 8000
```
In other words, there's an HTTP server running on port 8000 via `php -S` and it responds to curl. But, portfinder offers up its port as free.

But then, for the hell of it, I tried running a SimpleHTTPServer with Python.

This time, portfinder worked as expected:
```bash
➜ python -m'SimpleHTTPServer'
Serving HTTP on 0.0.0.0 port 8000 ...
```
(switch to another shell)
```bash
➜ curl -si http://127.0.0.1:8000 | head -1
HTTP/1.0 200 OK
➜ node -e 'require("portfinder").getPort(console.log)'
null 8001
```
So, I guess I'm wondering what are `adb` and `php` doing differently from the Python server that causes portfinder to not detect the occupied port?

eriktrom:
fixes #13

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -102,7 +102,7 @@ exports.getPort = function (options, callback) {
 
     var port = ports.pop();
     for (var i = 0; i < ports.length; i++) {
-      if (port !== ports[i]) { return exports.getPort({ host: options.hosts }, callback); }
+      if (port !== ports[i]) { return exports.getPort({ host: options.hosts, port: ports[i] }, callback); }
     }
 
     return callback(null, port);
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -4,12 +4,53 @@
  * (C) 2011, Charlie Robbins
  *
  */
- 
+
 var fs = require('fs'),
     net = require('net'),
     path = require('path'),
     async = require('async'),
     mkdirp = require('mkdirp').mkdirp;
+
+var internals = {};
+
+internals.testPort = function(options, callback) {
+  if (!callback) {
+    callback = options;
+    options = {};
+  }
+
+  options.port   = options.port   || exports.basePort;
+  options.host   = options.host   || null;
+  options.server = options.server || net.createServer(function () {
+    //
+    // Create an empty listener for the port testing server.
+    //
+  });
+
+  function onListen () {
+    options.server.removeListener('error', onError);
+    options.server.close();
+    callback(null, options.port);
+  }
+
+  function onError (err) {
+    options.server.removeListener('listening', onListen);
+
+    if (err.code !== 'EADDRINUSE' && err.code !== 'EACCES') {
+      return callback(err);
+    }
+
+    internals.testPort({
+      port: exports.nextPort(options.port),
+      host: options.host,
+      server: options.server
+    }, callback);
+  }
+
+  options.server.once('error', onError);
+  options.server.once('listening', onListen);
+  options.server.listen(options.port, options.host);
+};
 
 //
 // ### @basePort {Number}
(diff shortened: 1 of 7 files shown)


stefanpenner:
@eriktrom seems like CI is un-happy

lib/portfinder.js
```diff
@@ -4,37 +4,21 @@
...
-exports.basePath = '/tmp/portfinder'
-
-//
-// ### function getPort (options, callback)
-// #### @options {Object} Settings to use when finding the necessary port
-// #### @callback {function} Continuation to respond to when complete.
-// Responds with a unbound port on the current machine.
-//
-exports.getPort = function (options, callback) {
+internals.testPort = function(options, callback) {
```
indexzero: Why make this internal?

indexzero:
@eriktrom @stefanpenner this seems like a reasonable change, but I am not going to have a lot of time to look at it in the next week (work and travel). 

CI is indeed unhappy, and it's odd because I can reproduce passing tests on `4.2` locally, but they fail on Travis. If we can get the tests passing here and rebase off `master` (new TravisCI config targeting `node@4`, etc) then we're good to go here.

lib/portfinder.js
eriktrom: exporting it exposes the old api to the world that only checks the default port when calling net.createServer, which varies per system and leaves other interfaces//hosts on the same port open. 

The transplanted internals.testPort function is an exact copy of the original exports.getPort function so 2 other small goals(besides not exposing it) we're to: 1) Make code review easy 2) Allow the new getPort function easier to reason about.

I can inline it into the new getPort function if that's a better fit for the codebase. Not opposed.

indexzero:
@eriktrom any idea why the build is still failing?

eriktrom:
@indexzero - not completely - i did ask travis which ports we're open (https://github.com/eriktrom/ember-cli/commit/c21e6702af0fd35ea765a72d20b3367c58fb9234#commitcomment-15813337) - but on which interfaces i'm not clear - prior to this change we we're not testing ipv4 and ipv6 interfaces - my guess is that ::1 is not open for business on travis, any suggestions?

another thought - in that link port 32768 is the first available local port that travis says is open - opening 30,000 ports might be the issue as well as we start at 8000 (which works when `host: undefined` chooses the default interface and host) - we could start higher perhaps

last thought - i could stub/mock a bunch of stuff i suppose but seems like that would only test the stubbed world i would create

@stefanpenner if you have any ideas, drop them here :)

indexzero:
@eriktrom lets just start from port 32768 instead of 8000 then

test/helper.js
```diff
@@ -0,0 +1,32 @@
...
+
+  if (!next) {
+    server.listen(base, host);
+  } else {
+    server.listen(base, host, next);
+  }
+
+  server.on('error', function() {
+    server.close();
+  });
```
eriktrom: @indexzero - 

```
server.on('error', function() {
  server.close();
});
```

does not work in node v0.10 - neither does putting that inside on('listening') - otherwise this passes now (on 0.12 and node 4.x)

any thoughts?

eriktrom:
i will squash and rebase this when we're done here :) hahaha

indexzero:
@eriktrom I am OK making this `node >= 0.12` if we have to. That would simply make this `portfinder@1.0.0`

kevin-smets:
Eagerly awaiting this one, awesome work you guys!

eriktrom:
g2g

i version bumped package.json to 1.0.0, fyi (so need to release to npm)

indexzero:
Looks good. Thanks for your work!

stefanpenner:
:+1:

Issue #21, opened 2016-02-03
v1.0.0 portfinder no longer works


felicienfrancois:
Since v1.0.0, On windows 10 + nodejs 5.5,
portfinder give me ports that are in use by nodejs http server

indexzero:
@felicienfrancois pull-request welcome. In the meantime you can safely rely on `0.4.0`.

indexzero:
@eriktrom @stefanpenner looks like this is an unintended side effect of the changes in #20. If you have any cycles to help out it would be greatly appreciated.

eriktrom:
i'll take a look tonight 

@felicienfrancois - any more info you have is greatly appreciated. i dont have much windows usage experience. any steps i can take to repro? is your project open source? thanks

indexzero:
@eriktrom a great place to start might be to setup Appveyor on a fork. I'm also happy to make you a Github owner if you want to set it up on this repo directly.

stefanpenner:
I'll take a look at that PR and see if anything jumps out

eriktrom:
@stefanpenner - just started debugging this - looks like the first two test pass then we get `Error: listen EACCES`  - https://github.com/eriktrom/node-portfinder/pull/1

indexzero:
I believe this was fixed in #23. Was published in `portfinder@1.0.1`

dbuhrman:
I am still experiencing this issue in 1.0.3.  Portfinder returns a valid port on OSX, however, same app running on Windows 10 gives a port that is in use by another expressjs app.

eriktrom:
@dbuhrman - thanks for reporting - can you print your stack trace and versions of everything. Also can you run `npm i && npm test` inside node-portfinder itself on your machine, that would really help - our tests check for this scenario and run in appveyor but I'm not sure about anything on windows as I don't use it

If the tests pass then i need to borrow your laptop :)  (actually I have a branch that logs everything I can give you so I'll should only need your help twice at most, once if the tests fail and give us something good, thanks so much)

eriktrom:
@dbuhrman - also run this while your at it - while both are ports are open on different hosts  (if it works in windows, or if you know the equivalent)

```sh
netstat -an | grep THEPORT
```

dbuhrman:
[...]

PR #22, opened 2016-02-04
Add appveyor

eriktrom:
Add appveyor.yml to test against windows in CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: node_js
 node_js:
   - "0.12"
   - "4.2"
+  - "5.0"
 
 before_install:
   - travis_retry npm install -g npm@2.14.5
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,35 @@
+# http://www.appveyor.com/docs/appveyor-yml
+
+# Fix line endings in Windows. (runs before repo cloning)
+init:
+  - git config --global core.autocrlf true
+
+# Test against these versions of Node.js.
+environment:
+  matrix:
+    - nodejs_version: "0.12"
+    - nodejs_version: "4.2"
+    - nodejs_version: "5.0"
+
+# Install scripts. (runs after repo cloning)
+install:
+  - git rev-parse HEAD
+  # Get the latest stable version of Node 0.STABLE.latest
+  - ps: Install-Product node $env:nodejs_version
+  # Typical npm stuff.
+  - md C:\nc
+  - npm config set cache C:\nc
+  - npm version
+  - npm install
+
+# Post-install test scripts.
+test_script:
+  # Output useful info for debugging.
+  - npm version
+  - cmd: npm test
+
+# Don't actually build.
+build: off
+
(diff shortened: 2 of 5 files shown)


eriktrom:
@indexzero - do you want to add the appveyor integration on your side, no need for commit bit

once you do this will fail for the socket test but pass for the others (the socket support was not altered by #20) so could use your thoughts here

i'll push another commit that will show that it passes for non socket tests

fyi - @stefanpenner

eriktrom:
to summarize there are 2 issues we need to address:

- how to make the socket tests pass with appvoyer (non regression, just life bug)
- whether our tests cover the use case in #21, and if not, what tests to add to get the reported there

eriktrom:
@indexzero - this is closer - on appveyor we now only have one failing test, the same one as we have on travis above(im not done with the socket test, fyi, the cleanup using windows friendly node methods is WIP) and I believe #23 may have taken care #21 (although this is a hunch, but seems likely)

i'll be back at a later date :)

indexzero:
@eriktrom glad to hear you're making progress. Happy to give you commit bit here if you'd like. Sounds like you've got this situation under control :smile: :+1:

eriktrom:
@indexzero - I agree with a comment you made, we really need good tests. This is almost ready, but creating the servers within the socket tests throw - I console logged the error and made a vow (get it :) ) not to debug after midnight (given the typo i made that i think caused a couple of tickets recently, even if they did reveal deeper flaws) - im a bit wrapped up in some other things this week - can you look at the error on appveyor, where I put the console log and perhaps ask a windows buddy (i have none) whats up with the var server = net.createServer -> server.listen -> outputs this error:

```
err is %o { [Error: listen EACCES] code: 'EACCES', errno: 'EACCES', syscall: 'listen' }
C:\projects\node-portfinder\node_modules\vows\lib\vows.js:207
        console.log(console.result(results));
                ^
TypeError: undefined is not a function
    at process.<anonymous> (C:\projects\node-portfinder\node_modules\vows\lib\vows.js:207:17)
    at process.emit (events.js:107:17)
    at process.options.Emitter.emit (C:\projects\node-portfinder\node_modules\vows\lib\vows.js:240:24)
```

This error does not happen when i create non socket based servers in the other tests, so perhaps it has to do with the file system(that i dont understand in windows) that is making this throw or it could be something else, im not sure - its a bit hard to debug unless i can gain some context around the issue im trying to solve to make this error go away

thanks

indexzero:
@eriktrom sorry for the delay in getting back to you. Two points to help you out here:

1. There are subtle things about sockets on Windows that I never quite understood, but `forever` implements: https://github.com/foreverjs/forever/blob/master/lib/forever/worker.js#L141-L154
2. If the above can't sort it out then I am fine with skipping the socket-based tests on Windows since the vast majority of users do simple port-based listening. Let's just add a TODO to revisit it later on though.

eriktrom:
Bam - foreverjs hack worked first time, thanks @indexzero - we've got windows tests, w00t!

PR #23, opened 2016-02-04
Fix wrong duplicate removal

Loghorn:
This fixes the removal of `options.host` that is always happening because the loop starts from `0` instead of `1`. It also stops the check as soon as a duplicate is found.


eriktrom:
@Loghorn - thankyou - that was my bad

Loghorn:
you are welcome :smile:

indexzero:
Thank you for your contribution!

Issue #24, opened 2016-02-09
Does not work if the given base port is already taken


kevgo:
The test script below allocates port 3000, then calls portfinder with basePort set to 3000. 
With the old version 0.4.0 it works correctly, i.e. prints "1 2 3". 
With portfinder 1.0.0., it does not output "3", i.e. it never calls the callback given to `getPort`. 

```javascript
http = require('http');
portfinder = require('portfinder');
handler = function(_, res) {
  return res.end('existing server');
};

console.log(1);
server = http.createServer(handler).listen(3000, 'localhost', function() {
  console.log(2);
  portfinder.basePort = 3000;
  portfinder.getPort(function(err, port) {
    console.log(3);
  });
});
```

My setup:
* OS X 10.11.3
* Node v5.3.0

winfinit:
i've encountered similar issue, with this pull request https://github.com/indexzero/node-portfinder/pull/25 above is printing "3"

indexzero:
@kevgo does this work on `portfinder@0.4.0`?

kevgo:
Yes, it works in 0.4.0. This issue is new in 1.0.0.

winfinit:
@indexzero it would work, because of introduction of ::1, 0.0.0.0 and 127.0.0.1 test, basically previous bug was that if you start something on 0.0.0.0, and then test against 127.0.0.1, then same port would be reported as open, and that is actually correct in OSX, because 0.0.0.0 would listen as *.{port}, and * does not mean that listen on "EVERY" available interface, it is a default route for specified port, unless there was a bind on a specific device, so for example, this would successfully bind to all three interfaces:

```javascript
const http = require('http');

const port = process.argv[2] || 3000;

http.createServer( function(request, response) {
  response.writeHead(200, {'Content-Type': 'text/plain'});
  response.end('Hello World 0.0.0.0\n');
}).listen(port);

http.createServer( function(request, response) {
  response.writeHead(200, {'Content-Type': 'text/plain'});
  response.end('Hello World 127.0.0.1\n');
}).listen(port, "127.0.0.1");

http.createServer( function(request, response) {
  response.writeHead(200, {'Content-Type': 'text/plain'});
  response.end('Hello World ::1\n');
}).listen(port, "::1");
```

```
$ netstat -an |grep 3000
tcp6       0      0  ::1.3000               *.*                    LISTEN     
tcp4       0      0  127.0.0.1.3000         *.*                    LISTEN     
[...]

PR #25, opened 2016-02-11
fixed infinite loop when bind across local interfaces is different

winfinit:
currently in getPort, hosts ::1, 127.0.0.1, and 0.0.0.0 are treated the same, and if any of them are not, application falls into infinite loop, because "if" condition was testing just last and current port, while we have three ports, and if they happen not to match, recursion is triggered, however that recursion will always test "current" and last port, which always would return not true.

so with some extra console.log you would see this:

```
in testPortFn: ::1 undefined
in testPortFn: 0.0.0.0 undefined
in testPortFn: 127.0.0.1 undefined
in getPorts: [ 3000, 3001, 3001 ]
in testPortFn: ::1 3000
in testPortFn: 0.0.0.0 3000
in testPortFn: 127.0.0.1 3000
in getPorts: [ 3000, 3001, 3001 ]
in testPortFn: ::1 3000
in testPortFn: 0.0.0.0 3000
in testPortFn: 127.0.0.1 3000
in getPorts: [ 3000, 3001, 3001 ]
...
```

i've updated getPort function to test for unique items, and if more than one returned, it will sort that ports array, and trigger a new recursion against highest port

this issue would fix reported #24 issue.


lib/portfinder.js
```diff
@@ -77,39 +77,48 @@ exports.getPort = function (options, callback) {
...
-  for (var j = 0; j < hosts.length; j++) {
-    (function() {
-      var jj = j;
-      var testPortFn = function(callback) {
-        internals.testPort({ host: hosts[jj], port: options.port }, callback);
-      };
-
-      testPortFunctions.push(testPortFn);
-    }());
+  var findCommonOpenPort = function(portToTest) {
```
indexzero: style: `function findCommonOpenPort(portToTest) {`

lib/portfinder.js
```diff
@@ -77,39 +77,48 @@ exports.getPort = function (options, callback) {
...
+    var ports = [];
+    var error;
+    async.every(uniqueHosts, function(host, asyncCallback) {
```
indexzero: style: `async.every(uniqueHosts, function(host, next) {`

lib/portfinder.js
```diff
@@ -77,39 +77,48 @@ exports.getPort = function (options, callback) {
...
+      internals.testPort({"host": host, "port": portToTest}, function(err, port) {
+        if (err) {
+          error = err;
+          asyncCallback(err);
```
indexzero: `return next(err);`

indexzero:
Thank you for your contribution. This definitely needs new test coverage.

winfinit:
@indexzero since this behavior seems to be OS X specific, I am not exactly sure how to adequately structure such tests

eriktrom:
recursion against the highest port found for any interface already occurs here https://github.com/eriktrom/node-portfinder/blob/bd905520e538e6fe7a8cbed6fbae5b8d9b7b4ed2/lib/portfinder.js#L103-L106

the bug you reference perhaps lies here - when passing localhost as the options.host, we check the default interface for localhost, which is OS specific and possibly the same exact host that we have predefined in our array of hosts (say 127.0.0.1 or `::1`) thus causing an infinite loop

https://github.com/indexzero/node-portfinder/blob/master/lib/portfinder.js#L79-L88

when i walked way from the computer the day i wrote #20 my fear was this block of code, for a couple of reasons, but this one should have made my gut made my brain do a double take. I apologize. The question now is, if you pass in localhost, and localhost is 127.0.0.1 on one OS and `::1` on another, then what to do? I suppose remove localhost from the array of hosts - but that makes the assumption that localhost will only ever be `::1`, `0.0.0.0` or `::1`. If its anything else, the results of portfinder would potentially be wrong, although that case seems very rare.

@indexzero - would that cover the reported issue found in #24

eriktrom:
In other words, by passing localhost, we always recurse forever, if localhost is one the 3 predefined hosts that we already check for, because that will fail the old internals.testPort (aka, getPort in 0.4.0) check, so we keep recursing. I don't see any other issue though.

eriktrom:
@winfinit - to test my theory, try adding localhost to the array here https://github.com/indexzero/node-portfinder/blob/669ed062561cff1c522341977e9b9cdb1688d3cf/lib/portfinder.js#L80 on master and add your filter of that array(thankyou, thats the key i missed and #23 missed) and see if you still get the infinite loop

Besides the localhost case, the only other instances of infinite loops that can occur after that change would be if someone added their own host to /etc/hosts that points to any of our predefined hosts. To fix that we could simply run a internals.testPort against the provided options.host for each of our predefined hosts. 

For example i have machost.local aliased to 127.0.0.1 on my machine. If we we're to run internals.testPort, and pass in a callback that again calls internals.testPort, the first with options.port(machost.local in this example) and the second with 127.0.0.1(one of our predefined hosts), and we receive a `EADDRINUSE` error, then we know they are aliased as the same host, and can run `server.close();`, remove the provided options.host from the hosts array, break that little loop and move on knowing that we have no duplicate hosts (whether they're hostnames or addresses, etc). 

If that is what your code is doing, then ignore this comment. Otherwise, I must be missing something. 

Thanks for your contribution. Good work.

eriktrom:
also there was a typo on this line, should be options.host not options.hosts, effectively passing 'undefined' as the host, for the second recurse on up, and thus duplicating the default host for the OS

https://github.com/indexzero/node-portfinder/blob/669ed062561cff1c522341977e9b9cdb1688d3cf/lib/portfinder.js#L108

suppose thats what i get for working on this at 2am, my bad again

lib/portfinder.js
```diff
@@ -77,39 +77,48 @@ exports.getPort = function (options, callback) {
   }
 
   options.host = options.host || '::1';
-  var hosts = [options.host, '0.0.0.0', '127.0.0.1'];
+  var hosts = [options.host, '127.0.0.1', '0.0.0.0'];
```
eriktrom: add `::1` to this list if options.host is passed in, as that was a primary goal of #20 - thanks :)

winfinit:
@eriktrom to handle localhost and 127.0.0.1, we can just change async.each to async.eachLimit with limit of 1, that way both would successfully probe same port. Currently condition that you are describing is a race condition if I understand correctly, because two addresses would have to attempt to bind during test phase simultaneously (or very close to that), where one would report port as taken, which would cause second iteration to try next port, and then same condition would have to occur again and again.

if that is the case, then eachLimit would solve such problem

winfinit:
i mean to say not async.each, but async.every and async.everyLimit

eriktrom:
that sounds just dandy - you'll know if it works by run port finder w/ host: localhost - if that works, then much saved time (your good at that async library, I was very rusty :) )

eriktrom:
one more question while i've got you - when running say nginx with `listen *:8000;` - do we know what will happen? (in otherwords, what does "*" use as the default host+interface and are we prepared for that

if you don't know, no worries, we'll figure it out in due time

eriktrom:
answer to nginx question on my machine seems to be 0.0.0.0 so` async.everyLimit` should fix that as well :)   very nice hack

eriktrom:
so funny story - im terrible with advice - testPort already removes the listener and closes the server when a port is found - we can test the same host as many times as we want, we'll get the same answer, assuming we dont run in parallel 

everyLimit though will only run testPort up to the limit, which is 1 now - meaning we dont test the rest of the hosts - is that logic true?

if you want me to finish up here i dont mind, seems easy at this point, remove code, add test, done

(it all makes sense now!)

oh man what a saga, lol

winfinit:
@eriktrom limit would control concurrency, but will run all of the iterations, as long as an error doesn't return **false**, so in our instance, we would expect always to scan entire list of hosts, unless one of them threw an exception due to OS restrictions, i.e.: getPort would return **err**

i am currently attempting to write a test, i have one that is failing already, however issue that i am running into is that vows doesn't provide a timeout, and when i am adding a setTimeout before running getPort, it is never getting its tick, because getPort is never yielding when failure is exercised, and i would love it to actually return a failure, and not just hang...

current test that i have:
```javascript
var vows = require('vows'),
    assert = require('assert'),
    portfinder = require('../lib/portfinder'),
    http = require('http');

const host = "localhost";
var server, workingPort;


vows.describe('portfinder').addBatch({
  "When using portfinder module": {
    "get available port": {
      topic: function () {
		const cb = this.callback;
		portfinder.getPort(function(err, port) {
			assert.isFalse(!!err);
			workingPort = port;
			cb();
		});
      },
      "start a server on discovered port": {
        topic: function () {
			const cb = this.callback;
			server = http.createServer(function(){}).listen(workingPort, host, function() {
				cb();
			});	
        },
		"attempt to get run getPort with basePort of a server": {
			topic: function() {
			    console.log('getting a new port');
				const cb = this.callback;
				portfinder.basePort = workingPort;
				var timer = setTimeout(function() {
					cb("Unable to get port");	
				}, 2000); // 2 seconds
				process.nextTick(function() {
					portfinder.getPort({host: host}, function(err, port) {
						clearTimeout(timer);	
						cb(err);	
					});
				});
			},
        	"should not get an error": function (err) {
			  server.close();
        	  assert.isTrue(!err);
        	}
		}
      }
    }
  }
}).export(module);
```

eriktrom:
i just learned vows last nite - luckily @indexzero wrote it :) maybe he can help here - ping!

eriktrom:
perhaps though, you clear the timer before its ever put on the event loop - process.nextTick runs at the bottom of the event loop, timers are put onto the event loop after that 

var timer = setTimeout(function() { console.log('blah') }); process.nextTick(function() { clearTimeout(timer); console.log(timer) });

which is just like

var timer = setTimeout(function() { console.log('blah') }, 1000); setTimeout(function() { clearTimeout(timer); console.log(timer) }, 996);

(4ms between each event loop turn, process.nextTick runs immediately after sync code, perhaps this is why, perhaps its vows, sorry wish i knew)

also, you can pass the server and workingPort through via the callback instead of using the undefined vars at the top(which, if depending on order of callbacks, may still be undefined by the time the test for a topic runs)

maybe that helps if not, @indexzero will know - thats all i got :)

winfinit:
@indexzero I've added a test that would demonstrate behavior in #24, here is output from a test against 1.0.1
```
$ npm test

> portfinder@1.0.1 test /Users/rjurkov/workspace/old/node-portfinder
> vows test/*-test.js --spec

 
  ♢ portfinder 
  
  When using portfinder module ensure that infinite loop is not triggered
    ✗ should not get an error 
        » expected false, got true // port-finder-0-vs-127-test.js:43
  
  ♢ portfinder 
  
  When using portfinder module with 5 existing servers the getPorts() method with an argument of 3
    ✓ should respond with the first three available ports (32773, 32774, 32775)
  When using portfinder module with no existing servers the getPorts() method with an argument of 3
    ✓ should respond with the first three available ports (32768, 32769, 32770)
  
  ♢ portfinder 
  
  When using portfinder module with 5 existing servers the getPort() method
    ✓ should respond with the first free socket (test5.sock)
  When using portfinder module with no existing servers the getSocket() method with a directory that exists
    ✓ should respond with the first free socket (exists.sock)
  When using portfinder module with no existing servers the getSocket() method with a directory that doesnt exist
    ✓ should respond with the first free socket (test.sock)
  When the tests are over
    ✓ necessary cleanup should take place
  
  ♢ portfinder 
  
  When using portfinder module with 5 existing servers the getPort() method
    ✓ should respond with the first free port (32773)
  When using portfinder module with no existing servers the getPort() method
    ✓ should respond with the first free port (32768)
 
✗ Broken » 8 honored ∙ 1 broken (10.073s) 
  
npm ERR! Test failed.  See above for more details.
```

here is output with this pull request:
```
$ npm test

> portfinder@1.0.1 test /Users/rjurkov/workspace/node-portfinder
> vows test/*-test.js --spec

 
  ♢ portfinder 
  
  When using portfinder module ensure that infinite loop is not triggered
    ✓ should not get a timeout
  
  ♢ portfinder 
  
  When using portfinder module with 5 existing servers the getPorts() method with an argument of 3
    ✓ should respond with the first three available ports (32773, 32774, 32775)
  When using portfinder module with no existing servers the getPorts() method with an argument of 3
    ✓ should respond with the first three available ports (32768, 32769, 32770)
  
  ♢ portfinder 
  
  When using portfinder module with 5 existing servers the getPort() method
    ✓ should respond with the first free socket (test5.sock)
  When using portfinder module with no existing servers the getSocket() method with a directory that exists
    ✓ should respond with the first free socket (exists.sock)
  When using portfinder module with no existing servers the getSocket() method with a directory that doesnt exist
    ✓ should respond with the first free socket (test.sock)
  When the tests are over
    ✓ necessary cleanup should take place
  
  ♢ portfinder 
  
  When using portfinder module with 5 existing servers the getPort() method
    ✓ should respond with the first free port (32773)
  When using portfinder module with no existing servers the getPort() method
    ✓ should respond with the first free port (32768)
 
✓ OK » 9 honored (0.143s) 
```

I don't particularly like how test is structured, however when getPort is running, it never yields, so even if i would wrap it in nextTick, that would only yield ones, and when it goes to infinite loop... you are out of luck, so setTimeout is not an option.

so to combat this, i had to spawn offender in a separate process, so that way timer can get its tick.

I am thinking that it would be a good idea to maybe submit another pull request where between each port test, test port would yield, so that way one could implement some sort of a timeout, or other operations would get also a tick?

winfinit:
@indexzero any chance this can get merged?

indexzero:
For the record I'm fine with changing all of these tests to `mocha` if anyone is interested. I'm pretty swamped at work this week, but I should have time this weekend.

eriktrom:
I'll take a close look at this as well *before* the weekend then :)

thanks for the work @winfinit

indexzero:
@eriktrom you have a good grip of this stuff. You want commit bit and you can merge it in when you get time?

eriktrom:
sure that's fine

eriktrom:
@winfinit - do u have any interest in pairing on this. I see some things we can simplify a bit if you've got a moment in the next couple days - seems like an efficient way to ship this and raise its bus factor.

lib/portfinder.js
```diff
@@ -76,40 +76,52 @@ exports.getPort = function (options, callback) {
...
   // remove duplicate host arg if already in hosts array
-  for (var i = 1; i < hosts.length; i++) {
-    if (options.host === hosts[i]) {
-      hosts.shift();
-      break;
-    }
-  }
+  var uniqueHosts = hosts.filter(function(host, index) {
+    return hosts.indexOf(host) == index;
+  });
```
eriktrom: we can remove this, its okay if we have an array on non unique hosts - portfinder will close the server when it finds an open port for a given host - we have no way of preventing this anyway as a hostname can alias an ip, thus our goal at the end of this method will be to just ensure all the ports found match, if they don't recurse using the highest found port as the starting point. This makes this code more resilient and much shorter(which i'll point out in a sec) - it will also make testing easier - we can pass in duplicate hosts on purpose, and that's our proof that an infinite loop is not possible

lib/portfinder.js
```diff
@@ -76,40 +76,52 @@ exports.getPort = function (options, callback) {
...
-      var testPortFn = function(callback) {
-        internals.testPort({ host: hosts[jj], port: options.port }, callback);
-      };
-
-      testPortFunctions.push(testPortFn);
-    }());
+  function findCommonOpenPort(portToTest) {
+    var ports = [];
+    var error;
+    async.everyLimit(uniqueHosts, 1, function(host, next) {
```
eriktrom: so it does seem that asyncLimit does iterate through *until a falsy value is returned* - this would be fine except we need to handle an obscure error pointed out in https://github.com/indexzero/node-portfinder/issues/26

Basically not every user of every OS or corporate issued laptop, cloud booted server, etc - will have ipv6 open - so we need to catch the error that occurs when that interface is not available. That error is `EADDRNOTAVAIL` - there is always a chance I'm not fully correct about why or when this error will be thrown, but it will throw for some users - so we need to catch it and remove that host from the array when it is thrown, continue testing the rest of the hosts, and never test it again as it will always throw(i believe) for any port we try to lookup for it

aka, we catch this error, remove that host from the array, continue, never test that host again

this seems to affect the way this iterator is written in a couple of ways

lib/portfinder.js
```diff
@@ -76,40 +76,52 @@ exports.getPort = function (options, callback) {
...
+      } 
+
+      var uniquePorts = ports.filter(function(port, index) {
+        return ports.indexOf(port) == index;
+      });
+      if (uniquePorts.length > 1) {
+        // find highes port
+        uniquePorts.sort(function(a, b) {
+          return a - b;
+        });
```
eriktrom: lines 107-114 can be removed if we simply use an indexOf test - see if the port is already in the array before we push, if it is, don't push - when we're done, we just sort the array, test that the last is the same as the first, if not, use the last to recurse

lib/portfinder.js
```diff
@@ -76,40 +76,52 @@ exports.getPort = function (options, callback) {
...
-    if (err) { return callback(err); }
-
-    var port = ports.pop();
-    for (var i = 0; i < ports.length; i++) {
-      if (port !== ports[i]) { return exports.getPort({ host: options.hosts, port: ports[i] }, callback); }
-    }
-
-    return callback(null, port);
-  });
+  findCommonOpenPort(options.port || exports.basePort);
```
eriktrom: the internals.testPort function already starts at the port given, the exports.nextPort(i think is the name) simply adds +=1 from there, so we can remove this line (if you found differently, let me know, going off memory)

test/port-finder-0-vs-127-test.js
```diff
@@ -0,0 +1,47 @@
...
+      });  
+    });
+      },
+      "should not get a timeout": function (err, res) {
+        server.close();
+        assert.isFalse(!!res);
+      }
+    }
+  }
+}).export(module);
```
eriktrom: so a couple of things - most are my fault, this is pretty creative may have its merits - that said...

a) if we do keep it, can we rename the file to port-finder-integration-test.js
b) im not sure we need it - we have a similar setup already - we can add localhost to this array: https://github.com/indexzero/node-portfinder/blob/master/test/helper.js#L27 - i'd actually like to add an exact duplicate ip address there (you can add localhost as well, it should not matter, if our expected behavior works out right). This will test that we can pass any hostname, even if mapped to the same ip, more than once to our input, preventing pretty much all the bugs that have risen since 1.0

I'd also like to catch the error that only travis throws in that file, which is why I added these lines: https://github.com/indexzero/node-portfinder/blob/master/test/helper.js#L14-L16 - in this case, travis is just another user, he/she just happens to be setup differently - we should handle that error in our iterator(if it becomes hard, we'll file an issue for that, dont worry to much)

let me know your thoughts on all of this, thanks again for your help - if you want to hand over the work, i understand - that said, if you want to bang it out, go for it, you've done good work

eriktrom: Amending this comment(from previous inline comment on tests:)

> we can add localhost to this array: https://github.com/indexzero/node-portfinder/blob/master/test/helper.js#L27 - i'd actually like to add an exact duplicate ip address there (you can add localhost as well, it should not matter, if our expected behavior works out right).

instead, what i should have said is that we can create a test that looks like this:

```js
    "with 5 existing servers": {
      topic: function () {
        testHelper(servers, this.callback);
      },
      "the getPort({host: '127.0.0.1'}) method": {
        topic: function () {
          portfinder.getPort({ host: '127.0.0.1}, this.callback);
        },
        "should respond with the first free port (32773)": function (err, port) {
          assert.isTrue(!err);
          assert.equal(port, 32773);
        }
      }
    }
```

for say 127.0.0.1 and localhost

such a test falls falls nicely in line with [this current test](https://github.com/indexzero/node-portfinder/blob/master/test/port-finder-test.js#L23-L31), but test a specific host, something we we're not previously doing (thus our lack of coverage lies there) (technically we should perhaps have a 3rd test that tests a host and a port, and that the port starts iterating from the passed port, but we can make an issue for that if your short on time, no biggie)

winfinit:
@eriktrom do we have anything else left?

test/port-finder-integration-test.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+var vows = require('vows'),
+    assert = require('assert'),
+    portfinder = require('../lib/portfinder'),
+    child_process = require('child_process'),
+    path = require('path'),
+    http = require('http');
+
+const host = "localhost";
```
eriktrom: @winfinit - this pr looks good to go. Would you mind just changing this to var instead of const for jshint and doing a rebase. Then I pull the trigger. Thanks.

eriktrom:
Actually @winfinit - there is no jshint file and I just checked and const seems to work all the way back to node 10 so this is fine as is.

eriktrom:
@indexzero can you push to npm and ping me when you do. Thanks

eriktrom:
and bump version to 1.0.2

winfinit:
@indexzero any chance you can bump a version with this PR? I had to currently reference this dependency via git uri in mockgoose package, which is causing issues for some folks.

indexzero:
Sure thing folks. Been out of pocket the last couple of days starting my relocation to New York from California.

eriktrom:
Not to worry. Moving is all encompassing. Cali i think you'll dig though (my home state). Welcome to the west coast ;)

indexzero:
Other way around sir ;) ... relocating back to New York from California.

PR #27, opened 2016-02-25
Octal literals are not allowed in strict mode.

dcharbonnier:


lib/portfinder.js
```diff
@@ -155,7 +155,7 @@ exports.getSocket = function (options, callback) {
     options = {};
   }
 
-  options.mod  = options.mod    || 0755;
```
eriktrom: @dcharbonnier - thanks for the PR

That said, I *am* open to changing this, but given I had to hunt down the reasoning, I'm a little weary. The node source code says:

> // Maintainers, keep in mind that octal literals are not allowed
// in strict mode. Use the decimal value and add a comment with
// the octal value. Example:
//
//   var mode = 438; /* mode=0666 */

which makes sense as to why when I ran 

`parseInt(755, 8)` i get 493

So yes, ur change is correct. However, we can also use '0755' and node will take care of that for us, avoiding the confusion and preventing the error in strict mode. 

I suppose last note worth mentioning is that 'use strict' is not used within the source, so perhaps that's step 1. I'm willing to pull this in though if we simply use '0755' ATM, although we should make sure this behavior has not been removed in new versions of node (although that sounds very unlikely)

(ps - here is where they make that conversion for us in an old 0.x archive (found this on stack overflow)): https://github.com/nodejs/node-v0.x-archive/blob/c7f424e44b7e61b89f91cd344599c3d1cdfb88fe/lib/fs.js#L422-L423

Thanks!

eriktrom:
closing - we can re-open later if your still interested

dcharbonnier:
I think that being strict mode compatible is a good practice but I will just keep my fork, not a big deal.
parseInt(755, 8); is more explicit than 0755 that looks like a base 10 number.
I run node with --use-strict the code is executed faster, more clean and if I remember correctly provide few extra possibilities for ecmascript 6 in the nodejs 5 context

eriktrom:
@dcharbonnier - i actually agree with you. Do you want to add 'use strict' to the top of every file? There is a test that uses const so that should throw as reserved in 0.12 at least, but seems like everything else is strict compliant as is (and your right, lets enable it)

dcharbonnier:
done, thx!

dcharbonnier:
tested with :
- node/0.12.7
- node/0.10.32
- node/5.8.0

no test failures

eriktrom:
Awesome - thanks @dcharbonnier

Issue #26, opened 2016-02-17
1.0.1 breaks with EADDRNOTAVAIL


jeffcharles:
On updating to 1.0.1 of portfinder, I started getting the following error:

```
{ [Error: listen EADDRNOTAVAIL ::1:8000]
  code: 'EADDRNOTAVAIL',
  errno: 'EADDRNOTAVAIL',
  syscall: 'listen',
  address: '::1',
  port: 8000 }
```

Reverting to 1.0.0 fixes the issue.

eriktrom:
@jeffcharles - thanks for reporting that, that'll save some time when i fix the remaining bugs in 1.0

winfinit:
@eriktrom this one is pretty interesting... so we run into this exception when interface doesnt have ipv6 enabled... so we could add simple exception in a testPort to handle it, however "handling" becomes a bit opinionated, so for example, i would like to get a port for a host "::1", and interface is not enabled, i could report port is open (because ipv4 interfaces succeeded, and "EADDRNOTAVAIL" is ignored), however that is probably not what is expected... 

so what i am proposing is this: lets check if host is passed, and if host is an ip address, we check what type of ip it is, if it is ipv6, then we leave this exception as is, however if host is not passed, then we will ignore ::1 "EADDRNOTAVAIL", and report port as open.

what do you think?

winfinit:
@eriktrom actually another option is add a setting that will ignore ipv6 errors, or pop ::1 off of hosts array...

here is a version without extra option:

in getPort we could have this:
```javascript
  var hosts = ['::1', '127.0.0.1', '0.0.0.0'];
  if (options.host) {
    hosts.push(options.host);
    var addr = ipaddr.parse(options.host);
    if (ipaddr.isValid) {
        if (ipaddr.kind() === "ipv4") {
            internals.ignoreIPv6Error = true;
        }
    }
  }
```

and in testPort in onError add this:
```javascript
if (internals.ignoreIPv6Error === true && err.code === 'EADDRNOTAVAIL') {
        return callback(null, options.port);
}
```

eriktrom:
@winfinit - yes i agree - pop off the bad ones - related https://github.com/indexzero/node-portfinder/issues/26

eriktrom:
[...]

PR #28, opened 2016-03-10
Treat an EAFNOSUPPORT error as a common open port

corpulentcoffee:
On a system with the IPv6 stack turned off, testing against `::1` raises an `EAFNOSUPPORT` error and portfinder fails to return an open port.

Since the spirit of this is to find an open port "common" to all the `hosts` tested, I think an easy fix is to just treat `EAFNOSUPPORT` like "yes, this port is OK, since this host isn't even available on this system".

Thanks for the project!

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -36,6 +36,14 @@ internals.testPort = function(options, callback) {
   function onError (err) {
     options.server.removeListener('listening', onListen);
 
+    //
+    // If address is not supported on this system (e.g. `::1` on a system with
+    // the IPv6 stack disabled), treat as a "common" open port
+    //
+    if (err.code === 'EAFNOSUPPORT') {
+      return callback(null, options.port);
+    }
+
     if (err.code !== 'EADDRINUSE' && err.code !== 'EACCES') {
       return callback(err);
     }


indexzero:
@corpulentcoffee thank you for your contribution. This is an interesting bug indeed. I see the logic in what you're saying, but isn't this statement confusing to the user? 

> "yes, this port is OK, since this host isn't even available on this system".

My litmus test here is as follows: as a `node` developer I want the following to always end in a running `http` server or an error if the server can't listen:

``` js
var portfinder = require('portfinder');
var http = require('http');

portfinder.getPort(function (err, port) {
  if (err) { throw err; }

  http.createServer(function (req, res) {
    res.end(JSON.stringify('Hello.')):
  }).listen(port, function () {
    console.log('http server listening on %s', port);
  });
});
```

In this case, what happens to the `.listen('::1')` when IPv6 is disabled? Does it just silently fail?

corpulentcoffee:
Hey @indexzero:

On the master branch of portfinder, running that code produces this:

```
$ node test
/home/xxx/test.js:5
  if (err) { throw err; }
             ^

Error: listen EAFNOSUPPORT ::1:8000
    at Object.exports._errnoException (util.js:870:11)
    at exports._exceptionWithHostPort (util.js:893:20)
    at Server._listen2 (net.js:1218:19)
    at listen (net.js:1267:10)
    at net.js:1376:9
    at nextTickCallbackWith3Args (node.js:453:9)
    at process._tickCallback (node.js:359:17)
    at Function.Module.runMain (module.js:443:11)
    at startup (node.js:139:18)
    at node.js:968:3
```

With the patch to portfinder, that code can start a server:

```
$ node test
http server listening on 8000
^Z
[1]+  Stopped                 node test

program returned status code 148
$ bg 1
[1]+ node test &

$ curl http://localhost:8000
"Hello."
```

> In this case, what happens to the `.listen('::1')` when IPv6 is disabled? Does it just silently fail?

Not sure what you mean by this part (as `.listen('::1')` doesn't appear in that code you just included).

Cheers and thanks!

indexzero:
Hmmm ... I guess I need to investigate this more. If you could write a test along with this that would be super helpful.

corpulentcoffee:
Yes, I think short of running the test suite on an IPv6-disabled system, it might be possible to write a test for this case by specifying a custom server via `options.server` that returns the `EAFNOSUPPORT` error when fed an IPv6 address...?

I can investigate more.

eriktrom:
There is a related issue: https://github.com/indexzero/node-portfinder/issues/26

@indexzero - we can put the check in the iterator or where it's placed in this pr - we may have several cases as time goes by, its tough to preemptively know them all. It seems we should handle like @winfinit suggests in #26

eriktrom:
closed in favor of #29 - thanks @corpulentcoffee for reporting this issue.

PR #33, opened 2016-07-09
Handle MacOS 10.11.5+ interface changes and VM use case

eriktrom:
- allow portfinder.getPort for all ports w/ os.networkInterfaces().internal === true
- allow portfinder.getPort from netmask 255.255.255.0, which maps to 
  ip address given to default interface of machine (en0)
- allow portfinder.getPort to work from within a locally running VM assuming
  vnic0 or vnic1 use netmask 255.255.255.0 (aka, shared or bridged mode)
- handle 0.0.0.0 host that while not output by os.networkInterfaces() is
  nonetheless a bindable host that exists and must be handled accordingly

cc @indexzero or @stefanpenner - need at least one pair of eyeballs on this one. Thanks!

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -8,6 +8,7 @@
 "use strict";
 
 var fs = require('fs'),
+    os = require('os'),
     net = require('net'),
     path = require('path'),
     async = require('async'),
@@ -72,8 +73,6 @@ exports.basePath = '/tmp/portfinder'
 // #### @callback {function} Continuation to respond to when complete.
 // Responds with a unbound port on the current machine.
 //
-
-var defaultHosts = ['::1', '127.0.0.1', '0.0.0.0'];
 exports.getPort = function (options, callback) {
   if (!callback) {
     callback = options;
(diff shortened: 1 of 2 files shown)


eriktrom:
cc @dbuhrman - does this fix your use case? (if you have time to check thanks, if not, no biggie)

stefanpenner:
@eriktrom it seems good, but without tests its hard to see exactly what use-case this is addressing so Im not sure my feedback is that valuable as I haven't played with 10.11.x yet myself.

eriktrom:
@stefanpenner - thanks for the review - the tests did get an update - they iterate over the `exports._defaultHosts` now, so they reflect the current changes. 

Gonna merge - if we find issues, I'll make another PR

Thanks again for the review

eriktrom:
@indexzero - can you make a release to npm? Gratzi!

Turbo87:
@indexzero it would also be awesome if you could add @eriktrom as a collaborator on NPM to reduce the bus factor of this project

indexzero:
@eriktrom @Turbo87 done. Feel free to publish at will. Good job here 👍

eriktrom:
Thanks @indexzero 👍

PR #37, opened 2016-08-16
Catch eaddrnotavail

eriktrom:
Catch EADDRNOTAVAIL to filter out non-bindable hosts before iterating through ports

The general idea here is to filter out non-bindable hosts prior to looking for an open port so that we don't try to guess what hosts are bindable. 

Specifically worth noting though is that **the way this works does not actually filter all hosts before trying to find a port, it interleaves the two**. In practice however, the interfaces produced by `os.networkInterfaces()` are ordered from least to most likely to 'be bindable' so for most imaginable use cases, the port found will be 'the next open port for all bindable hosts' on the machine. There are cases however, for example when binding to an address from the "host machine" -> VM, where catching `EADDRNOTAVAIL` will (for example) discard 2 potentially working ports while also discarding the hosts that raised `EADDRNOTAVAIL`. Given the tradeoff between simple and 'less simple' this PR chose the simple path with the goal being to isolate the simplest change. This can be iterated on once we know this is the correct fix, although there is likely little to gain.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,3 +15,4 @@ script:
 notifications:
   email:
     - charlie.robbins@gmail.com
+    - erik.trom.github@gmail.com
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_install:
   - travis_retry npm install
 
 script:
-  - npm test
+  - DEBUG=* npm test
 
 notifications:
   email:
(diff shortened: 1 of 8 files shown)


eriktrom:
To he who may come after - forgive the long commit line, but it may save much debugging later to keep all commits around. Yours truly, erik.

PR #41, opened 2016-10-10
Don't bail when os.networkInterfaces is not supported.

kayakyakr:
Usage of non-supported operation under Windows Subsystem for Linux

os.networkInterfaces(), line 395, throws a not supported error on WSL (bash for windows). 

The operation will be supported at some point, but for now, I would want it to not crash and just return default values.

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -395,6 +395,7 @@ exports._defaultHosts = (function() {
   var interfaces = {};
   try{
     interfaces = os.networkInterfaces();
+  }
   catch(){
     // swallow error because we're just going to use defaults
   }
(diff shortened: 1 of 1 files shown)


kayakyakr:
Relevant WSL issue: https://github.com/Microsoft/BashOnWindows/issues/468#issuecomment-251298565

You may just wait until it is resolved in WSL itself, as a fix is in progress. Felt like there should be some documentation in place.

eriktrom:
@kayakyakr - thanks for the PR - let me try this out as well

eriktrom:
Thanks again @kayakyakr - I integrated your changes into #42 (along with changes for windows 10 native shell)

eriktrom:
@kayakyakr - thanks again!

Issue #39, opened 2016-09-05
Port is being re-bound under windows 10 (native shell) preventing portfinder logic from working


dharmax:
However, this simple code does work:
`
export function findFreePort() {
    return new Promise((resolve, reject) => {
        var server = createServer()
            , port = 0
        server.on("listening", function () {
            port = server.address().port
            server.close()
        })
        server.on("close", function () {
            resolve(port)
        })
        server.on('error', function (err) {
            reject(err)
        })
        server.listen(0, '127.0.0.1')
    })
}
`

eriktrom:
howdy @dharmax - can you provide more info? 

What do you mean by

> it always gives me 800

do you mean `8000`? The base port starts at 8000 here: https://github.com/indexzero/node-portfinder/blob/39e39603ec1f5dd27d5fafa9ea1dc7d2e5a558fd/lib/portfinder.js#L73

If you give me some more detail I'm more than glad to help work through this with you.

Thanks!

dharmax:
Sorry, 8000, yeah. 
Even if it already in use, of course.

eriktrom:
Thanks @dharmax for confirming - which version of portfinder are you running?

If your running the latest version, do mind setting this env var before you start your process:

`DEBUG=portfinder*`

e.g.,:

`DEBUG=portfinder* node index.js`

If you have a chance, paste that into a gist and drop me the link - I'll figure it out from there. 

Thanks again

noahzark:
@eriktrom @indexzero 
Hi,

 I also have the same probelm on Windows 10 (Version 10.0.10586) development machine, but works well on Linux machines. Seems windows 10 is allowing http.createServer(express()) and net.createServer listening on the same port with the same host (though the net server isn't working).

The screenshots are here:
http://imgur.com/a/libNy

Here are the steps to reproduce:

1. I have an express http server listening on 10004
http://imgur.com/laHP9rP

2. When I set basePort to 10004 and call portfinder.getPort(), it doesn't throw an error. Instead, it fires a listening event and calls the onListen callback
http://imgur.com/HbXst7C

3. When breaking on line 43 of portfinder, I tried the "netstat -an | grep 10004", and it shows two programs are listening on 0.0.0.0:10004
http://imgur.com/b8RGdDU

>   TCP    0.0.0.0:10004          0.0.0.0:0              LISTENING
>   TCP    0.0.0.0:10004          0.0.0.0:0              LISTENING
[...]

PR #42, opened 2016-10-17
Fix windows 10 support

eriktrom:
- Windows 10 native shell(#39) was rebinding to already bound ports. We
  now send a message from a socket client -> server -> client to ensure
  a host + port are actually open
- Windows 10 bash shell falls back to using 0.0.0.0 as the only host we
  check in order to prevent throwing a syscall error as being tracked here:
  https://github.com/Microsoft/BashOnWindows/issues/468

fixes: #39
closes: #41
fixes: https://github.com/ember-cli/ember-cli/issues/6338

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -16,6 +16,7 @@ var fs = require('fs'),
     mkdirp = require('mkdirp').mkdirp;
 
 var debugTestPort = debug('portfinder:testPort'),
+    debugTryConnection = debug('portfinder:tryConnection'),
     debugGetPort = debug('portfinder:getPort'),
     debugDefaultHosts = debug('portfinder:defaultHosts');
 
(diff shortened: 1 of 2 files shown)


lib/portfinder.js
```diff
@@ -29,28 +30,48 @@ internals.testPort = function(options, callback) {
...
+        options.server.close();
+        client.end();
+      } else {
+        options.server.emit('error', new Error(internals.windows10BindError));
+      }
+    });
+
+    client.on('end', function() {
+      callback(null, options.port);
+    });
```
eriktrom: future portfinder 2.0 partial code - includes fix for windows 10 native shell port rebinding and macbookpro with touchbar - and any other future issues that arise when **listening** on a port is not adequate - to use this portfinder needs to catch any OS errors that occur from trying to pipe data through a protected interface, such a bluetooth(and likely macbookpro touchbar) both which are interfaces returned by `os.networkInterfaces()`

PR #48, opened 2017-01-17
Typescript definition

jucrouzet:
Adds typescript definition to use module in .ts files natively.

### DIFF
--- a/lib/portfinder.d.ts
+++ b/lib/portfinder.d.ts
@@ -1,0 +1,17 @@
+/**
+ * portfinder.js typescript definitions.
+ *
+ * (C) 2011, Charlie Robbins
+ */
+
+declare type PortfinderHandler = (err: Error, port: number) => void;
+
+/**
+ * The lowest port to begin any port search from.
+ */
+export declare let basePort: number;
+
+/**
+ * Responds with a unbound port on the current machine.
+ */
+export declare function getPort(callback: PortfinderHandler): void;
--- a/lib/portfinder.d.ts
+++ b/lib/portfinder.d.ts
@@ -6,6 +6,17 @@
 
 declare type PortfinderHandler = (err: Error, port: number) => void;
 
+declare interface PortFinderOptions{
+  /**
+   * Host to find available port on.
+   */
+  host: string;
+  /**
+   * Minimum port (takes precedence over `basePort`).
+   */
+  port: number;
+}
+
 /**
  * The lowest port to begin any port search from.
  */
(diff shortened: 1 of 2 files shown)


jucrouzet:
Hi @indexzero (or any maintainer),
This PR only adds a definition for TS, it does not expose any new features or change an existing one, the goal is only to avoid an external dep in DefinitivelyTyped, is this possible for you to merge it quick ?
If you don't want to handle TS, let me know, I'll create a DefinitivelyTyped def, no problem.

Thanks !

lib/portfinder.d.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * The lowest port to begin any port search from.
+ */
+export declare let basePort: number;
+
+/**
+ * Responds with a unbound port on the current machine.
+ */
+export declare function getPort(callback: PortfinderHandler): void;
```
eriktrom: @jucrouzet - thanks a bunch - `getPort` actually takes an optional options object and a callback. The options are `host: string` and `port: number` - do u mind adding those? (https://github.com/indexzero/node-portfinder/blob/c5dd2cc76bd26d4771cee9e0637dfa27a7f89f24/lib/portfinder.js#L25-L28)

also - if u pass a port option, it takes precedence over the `exports.basePort` setting - https://github.com/indexzero/node-portfinder/blob/c5dd2cc76bd26d4771cee9e0637dfa27a7f89f24/lib/portfinder.js#L30
(not sure how u declare that in TS, if not possible, no worries)

mostly, do u mind adding the options object?

then i'll merge this in, thanks again

jucrouzet: Ok, thought that it was not "public" as it was not in the README.
The fact that the `port` option takes precedence over `basePort` has no implication in definition but I add it on comment for reference.

eriktrom: awesome thank you

yeah the port and host are public, README needs a lil update

(technically u can pass a server option too, that is NOT public :) )

again thanks for the commit

jucrouzet: YW
Embed TS definitions are a lot safer than DefinitivelyTyped ones ...

jucrouzet:
Sorry @eriktrom I went a little too fast on this one, got to make options optional, can you reopen the PR ?

eriktrom:
@jucrouzet - no problem - b/c it was just type declarations i pushed to npm already, open another pr with the fixes - thanks for taking the time to add that - i'll merge and re-publish once ur ready

PR #50, opened 2017-01-26
Promise support

jucrouzet:
Adds a `getPortPromise()` method that returns a Promise to the next free port : 

```js
const portfinder = require('portfinder');

portfinder.getPortPromise()
  .then((port) => {
      //
      // `port` is guaranteed to be a free port
      // in this scope.
      //
  })
  .catch((err) => {
      //
      // Could not get a free port, `err` contains the reason.
      //
  });
```

As  `getPortPromise()` is only a wrapper around `getPort()` I only added a simple unit test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,26 @@ The `portfinder` module has a simple interface:
   });
 ```
 
+Or with promise (if Promise are supported) :
+
+``` js
+  const portfinder = require('portfinder');
+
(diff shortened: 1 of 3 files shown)


jucrouzet:
There seems to be a sync problem with opened ports on CI tests, I close this one and I'll open a new one when it's ok

PR #51, opened 2017-01-26
Promise support

jucrouzet:
Adds `getPortPromise()` which returns a Promise to the next available port : 

``` js
  const portfinder = require('portfinder');

  portfinder.getPortPromise()
    .then((port) => {
        //
        // `port` is guaranteed to be a free port
        // in this scope.
        //
    })
    .catch((err) => {
        //
        // Could not get a free port, `err` contains the reason.
        //
    });
```

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -171,20 +171,22 @@ exports.getPort = function (options, callback) {
 // #### @options {Object} Settings to use when finding the necessary port
 // Responds a promise to an unbound port on the current machine.
 //
-if (typeof Promise === 'function') {
-  exports.getPortPromise = function (options) {
-    if (!options) {
-      options = {};
-    }
-    return new Promise(function(resolve, reject) {
-      exports.getPort(options, function(err, port) {
-        if (err) {
-          return reject(err);
-        }
-        resolve(port);
-      });
+exports.getPortPromise = function (options) {
+  if (typeof Promise !== 'function') {
+    throw Error('Native promise support is not available in this version of node.' +
+      'Please install a polyfill and assign Promise to global.Promise before calling this method');
+  }
+  if (!options) {
+    options = {};
+  }
+  return new Promise(function(resolve, reject) {
+    exports.getPort(options, function(err, port) {
+      if (err) {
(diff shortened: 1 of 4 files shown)


eriktrom:
thanks @jucrouzet - I'll take a close loo

i really appreciate the pull request - if it takes me more than a day or so for me to review its cause there are a number of downstream deps and we just have to be careful due to that

lib/portfinder.js
```diff
@@ -167,6 +167,27 @@ exports.getPort = function (options, callback) {
 };
 
 //
+// ### function getPortPromise (options)
+// #### @options {Object} Settings to use when finding the necessary port
+// Responds a promise to an unbound port on the current machine.
+//
+if (typeof Promise === 'function') {
```
eriktrom: Lets export this api regardless of whether native promise support is missing. If Promise is not available, lets still export the method and `throw Error('Native promise support is not available in this version of node. Please install a polyfill and assign Promise to global.Promise before calling this method`

or something like that

tl;dr - i dont want to conditionally export a public api method without a nice warning that the runtime does not support it, otherwise the error would be `portfinder.getPortPromise is undefined/not a function` or whatever - u get the drift

we could include bluebird but i don't want to bloat the size of the lib either

Feel free to notify the user w/ a message u see fit, including relevant details. Maybe update the README.md to mention this too.

thanks again - this looks good otherwise

jucrouzet:
Done.
I'm also not for including Bluebird, for a LOT of reason.

jucrouzet:
For the downstream deps, that's what came first to my mind, that's why I added a new method instead of modifying the `getPort()` method to return a promise when called with no callback.
BTW, I forgot  the TS definition, done.

jucrouzet:
Hi,
I've check all my projects using portfinder (~30).
They're all doing good with this PR, (some uses it bare, some use it with bluebird's promisify).

Let me know if there is something to change.

eriktrom:
@jucrouzet - yes this is fine - thanks for adding it to the README and fixing up the error

eriktrom:
published to npm

PR #53, opened 2017-03-08
Embrace Errors

eriktrom:

PR #56, opened 2017-05-03
Make test close listening sockets instead of just removing them

tomhughes:
Without this I see the tests (with node 6.10.2) completing but then just hanging because the servers are still listening on the sockets.


eriktrom:
@tomhughes - thank you - and makes sense, totally needed. Awesome. 

I'll likely cut a release with this included when #53 lands, which is soon, fyi.

PR #58, opened 2017-12-10
Update portfinder to convert option.port to number

debudda:
Hi! Got a strange error while working on an angular app. For some reason, when I try to get the server up it throws the error below.
Obviously, when it receives '8080', everything crashes and according to exports.basePort options.port always should be of a number type.

~ ng s
│net.js:1487
│      throw new errors.RangeError('ERR_SOCKET_BAD_PORT', options.port);
│      ^
│
│RangeError [ERR_SOCKET_BAD_PORT]: Port should be > 0 and < 65536. Received 80801.
│    at Server.listen (net.js:1487:13)
│    at Object.internals.testPort (/<projectPath>/node_modules/portfinder/lib/portfinder.js:69:18
│)
│    at Server.onError (/<projectPath>/node_modules/portfinder/lib/portfinder.js:60:15)
│    at Object.onceWrapper (events.js:254:19)
│    at Server.emit (events.js:159:13)
│    at emitErrorNT (net.js:1387:8)
│    at _combinedTickCallback (internal/process/next_tick.js:138:11)
│    at process._tickCallback (internal/process/next_tick.js:180:9)

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -27,7 +27,7 @@ internals.testPort = function(options, callback) {
     options = {};
   }
 
-  options.port   = +options.port   || exports.basePort;
+  options.port   = Number(options.port) || Number(exports.basePort);
   options.host   = options.host    || null;
   options.server = options.server  || net.createServer(function () {
     //
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -27,9 +27,9 @@ internals.testPort = function(options, callback) {
     options = {};
   }
 
-  options.port   = options.port   || exports.basePort;
-  options.host   = options.host   || null;
-  options.server = options.server || net.createServer(function () {
+  options.port   = +options.port   || exports.basePort;
+  options.host   = options.host    || null;
+  options.server = options.server  || net.createServer(function () {
     //
     // Create an empty listener for the port testing server.
     //


lib/portfinder.js
```diff
@@ -27,9 +27,9 @@ internals.testPort = function(options, callback) {
     options = {};
   }
 
-  options.port   = options.port   || exports.basePort;
-  options.host   = options.host   || null;
-  options.server = options.server || net.createServer(function () {
+  options.port   = +options.port   || exports.basePort;
```
eriktrom: wow - yeah this is really a good find - odd thing is that your the first to catch it. most libs must coerce their cli args to a type before passing the port as an option.

do u mind using `Number(options.port) || Number(exports.basePort)` - aka, wrap both - u can set the default, surely someone will set it as a string someday - and the `+` coercion is less obvious when reading the code...

will merge immediately and release after, thanks again

eriktrom:
also 

```
  When using portfinder module ensure that infinite loop is not triggered
    ✗ should not get a timeout 
        » expected false, got true // port-finder-z-integration-test.js:50
```

that test is run last and it has timeout on it - increase it if this passes locally for u - its kind of an odd test but is meant to catch e2e issues should they arise. travis must under load for that to fail (or its catching a regression, i would need look further, but its pretty obvious when ur down there. ping me with any q's - i can take over at any point, if u want, just let me know

u could add a test for the 8080 issue if u have the time budget too, i may after, do as much or as little to this pr as u want, i'll come back around in a week and fix and merge if not.. fyi

debudda:
Hi! Thanks for your advice, I'll probably have some time and follow your recommendations on a next week. I looked on the test complaint at that time, but didn't come up with a quick solution.

eriktrom:
Sounds good. 

For the test just bump the timeout. I’ve bumped it once before. It just ensures changes don’t cause an infinite loop. I think it’s 2000ms atm. Even 4000 is fine. It just needs to finish.

Thanks again

eriktrom:
Awesome. Thanks again.

PR #63, opened 2018-07-30
Adds stopPort to getPort Api

Prithvirajbilla:
- Adds stopPort option to getPort Api
  - This enables getPort api to find open port in the range basePort - stopPort/
- also includes some code boyscouting

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -27,8 +27,6 @@ internals.testPort = function(options, callback) {
     options = {};
   }
 
-  options.port   = Number(options.port) || Number(exports.basePort);
-  options.host   = options.host    || null;
   options.server = options.server  || net.createServer(function () {
     //
     // Create an empty listener for the port testing server.
@@ -73,6 +71,12 @@ internals.testPort = function(options, callback) {
 exports.basePort = 8000;
 
 //
+// ### @highestPort {Number}
+// Largest port number is an unsigned short 2**16 -1=65335
+//
+exports.highestPort = 65535;
+
+//
 // ### @basePath {string}
 // Default path to begin any socket search from
 //
@@ -88,10 +92,16 @@ exports.getPort = function (options, callback) {
   if (!callback) {
     callback = options;
     options = {};
+
   }
 
-  // Largest port number is an unsigned short 2**16 -1=65335
-  options.stopPort = Number(options.stopPort) || 65535
+  options.port   = Number(options.port) || Number(exports.basePort);
+  options.host   = options.host    || null;
+  options.stopPort = Number(options.stopPort) || Number(exports.highestPort)
+
+  if(!options.startPort) {
+    options.startPort = options.port
+  }
 
   if (options.host) {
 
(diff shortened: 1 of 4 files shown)


test/port-finder-test.js
```diff
@@ -46,6 +46,35 @@ vows.describe('portfinder').addBatch({
...
+          // stopPort: 32774 is greater than available port 32773 (32768 + 5)
+          setTimeout(function() {
+            portfinder.getPort({ stopPort: 32774 }, this.callback);
+          }.bind(this), 3000); //wait for cleanup of bound hosts.
+        },
+        "should respond with the first free port (32773) less than provided stopPort": function(err, port) {
+          if (err) { debugVows(err); }
+          assert.isTrue(!err);
+          assert.equal(port, 32773);
+        }
```
eriktrom: i think the socket is not fully closed by the time these are running for port 32773 - that's what those setTimeouts of 3 seconds are for -- its hacky i know - if you know vows and have a better way, by all means... 

it passes locally i assume?

eriktrom: also check https://github.com/indexzero/node-portfinder/blob/master/test/helper.js - we start some httpServer's in there, might be messing with your stopPort

I can't recall, but will look deeper before we merge for sure

lib/portfinder.js
```diff
@@ -88,6 +92,15 @@ exports.getPort = function (options, callback) {
...
     callback = options;
     options = {};
+
+  }
+
+  options.port   = Number(options.port) || Number(exports.basePort);
+  options.host   = options.host    || null;
+  options.stopPort = Number(options.stopPort) || Number(exports.highestPort)
+
+  if(!options.startPort) {
```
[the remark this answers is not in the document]
Prithvirajbilla: a) Done. 
b) Done. But question. If a user provides a port less than 1024, then we will not able to bound to them and eventually  program will return a port greater than 1024. So is this really needed?
c) Add one more check to guard against startPort is less than stopPort. 

Thanks for quick review.

test/port-finder-test.js
Prithvirajbilla: Yes, they are all passing locally.

Prithvirajbilla: Yeah. you are right. All the tests were run asynchronously when test servers are spun up and there is a race condition in the tests. I think the way to fix this is to have all the tests spun their own test services which are not colliding. I'd like to do this as a followup.

eriktrom: >  I think the way to fix this is to have all the tests spun their own test services which are not colliding. I'd like to do this as a followup.

agreed, and sounds awesome, thanks

lib/portfinder.js
```diff
@@ -88,6 +92,22 @@ exports.getPort = function (options, callback) {
...
+
+  }
+
+  options.port   = Number(options.port) || Number(exports.basePort);
+  options.host   = options.host    || null;
+  options.stopPort = Number(options.stopPort) || Number(exports.highestPort);
+
+  if(!options.startPort) {
+    options.startPort = Number(options.port);
+    if(options.startPort < 1024) {
```
eriktrom: https://github.com/indexzero/node-portfinder/issues/65 filed due to the above line. @Prithvirajbilla do u mind making a quick pr - or i can but starting work in like.. now so would be in a few hours

eriktrom: https://github.com/indexzero/node-portfinder/pull/66

Prithvirajbilla: I saw your PR. Thanks for getting to this quickly.  I will clean this up and will add a test for this special case (port 80) when I get time.

eriktrom: LOL - yeah you'd be surprised how many upstream libs use this thing :) no worries, happens.

PR #68, opened 2018-08-20
Makes test run sequentially instead of asynchronous.

Prithvirajbilla:
Also cleans up a comment.
Adds a regression test for indexzero/node-portfinder#65

We are better off running tests _sequentially_ than _asynchronous with setTimeout_.  The way to do this in vows is by creating a new batch.  There is some violation of DRY (compared to previous) with having same code in different places, But i wouldn't mind violating DRY in tests for more clarity and deterministic tests. 

Thanks.


eriktrom:
Very awesome, thanks!

Prithvirajbilla:
Thanks for reviewing this quickly. 

Let me know if you have any views on moving to widely used test framework like jest, etc.  I can give that a try this weekend.

PR #69, opened 2018-09-30
Upgrade vows to avoid security warning.

kylev:
Current npm will warn or error when asked to install dev dependencies
due to an outdated version of vows which has a potential DoS flaw.

https://nodesecurity.io/advisories/118


eriktrom:
@kylev - gratzi!

PR #72, opened 2018-11-30
Include getPorts in the typescript definition

rgbkrk:


eriktrom:
@rgbkrk - thanks!

rgbkrk:
Thanks for making a release so quick for this!

PR #75, opened 2019-01-14
fixed next port when port is not a number.

dingyi1993:
When I pass a string port,'nextPort' function will get a unexpected result.


eriktrom:
> When I pass a string port,'nextPort' function will get a unexpected result.

this has been fixed via #82 - we no longer call nextPort from getPort. we do call it from getPorts, but in that case, the initial port is passed to getPort which always returns a number.

PR #77, opened 2019-05-29
Avoid crashing if all ports above the start port are unavailable

pimterry:
Right now, calling `getPort` when _every_ port above the given port is in use will throw:

```
RangeError [ERR_SOCKET_BAD_PORT]: Port should be >= 0 and < 65536. Received 65536.
```

This isn't passed to the callback - it's thrown directly by node. Because this isn't handled at all that will usually crash the whole node process.

To easily reproduce this, start a server on port 65535 (the maximum port), and then run `portfinder.getPortPromise({ port: 65535 })`.

This happens because `testPort` is unbounded, and constantly tries ports until it finds an available one, without checking whether the port to test is valid, so when 65535 is in use it tries to test 65536, which isn't a valid port.

This PR makes `testPort` instead check whether the next port is valid before testing it, so that it fails cleanly if there are no ports available.

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -52,8 +52,14 @@ internals.testPort = function(options, callback) {
       return callback(err);
     }
 
+    var nextPort = exports.nextPort(options.port);
+
+    if (nextPort > exports.highestPort) {
+      return callback(new Error('No open ports available'));
+    }
+
     internals.testPort({
-      port: exports.nextPort(options.port),
+      port: nextPort,
       host: options.host,
       server: options.server
     }, callback);
--- a/test/helper.js
+++ b/test/helper.js
@@ -20,11 +20,18 @@ function createServer(base, host, next) {
   return server;
 }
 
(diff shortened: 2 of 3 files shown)


pimterry:
@eriktrom anything I could do to help get this merged?

eriktrom:
@pimterry - this looks good, sorry for the delay - this is actually useful for macbook pro's that have a touch bar as well, b/c apple randomly chooses a port between 40,000 and 65xxxx to send socket messages to and from the touch bar.

Thanks for the pr!

pimterry:
Thanks @eriktrom! Looking forward to the next release with this.

One thing to note, it looks like the merge failed one of its builds on Appveyor: https://ci.appveyor.com/project/indexzero/node-portfinder/builds/25912795. Seems to be ECONNRESET errors from npm during install, so it should pass again if you rerun it, just fyi.

eriktrom:
@pimterry - will check, thanks for noticing. Will publish tonight too. 

👌

Issue #79, opened 2019-08-02
Replace sequential scan with random search


BarthesSimpson:
Instead of scanning from the base port to the highest port, a randomized search would be more robust. Currently, if I am launching several servers in quick succession, there can be port collisions because one server may not have finished binding to the port discovered by portfinder before the next call to getPort, which then considers the same port to still be unbound.

eriktrom:
> Currently, if I am launching several servers in quick succession, there can be port collisions because one server may not have finished binding to the port discovered by portfinder before the next call to getPort, which then considers the same port to still be unbound.

Yes, this an issue I am aware of. Randomizing the port scanning is an interesting idea -- open to a PR on this, thanks!

BarthesSimpson:
(Sorry for the triplicate PR - my GitHub chops are a bit rusty)

eriktrom:
fixed by #82

PR #80, opened 2019-08-10
Randomize the order in which ports are tested for availability

BarthesSimpson:
Addresses https://github.com/http-party/node-portfinder/issues/79 by creating an array of candidate ports to test and randomly shuffling it. The remaining logic is unchanged, except that the check for whether a tested port exceeds the global `highestPort` is moved from `testPort` to `getPort` (i.e. we exclude any port higher than `highestPort` from the initial list of candidate ports).

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -33,18 +33,21 @@ internals.testPort = function(options, callback) {
     //
   });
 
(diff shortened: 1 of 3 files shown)


BarthesSimpson:
Closing until I figure out why the build is breaking.

Issue #85, opened 2019-08-17
Breaking change in 1.0.22 if port is > 40000 and stopPort is not set


jelhan:
Default highest port was reduced in https://github.com/http-party/node-portfinder/pull/82 from `65535` to `40000`. This was discussed in https://github.com/http-party/node-portfinder/pull/82#issuecomment-520199847 and explicitly stated to be done to _not_ introduce a breaking change. But it has indeed if `getPort()` is called with `port > 40000` and without `stopPort` and without overriding `portfinder.highestPort` before. E.g. this usage:

```js
portfinder.getPort({
    port: 49741
}, callback);
```

throws

> Provided options.stopPort(40000is less than options.startPort (49741)

Noticed this breaking change by failing tests that depend on https://github.com/tomdale/ember-cli-addon-tests/issues/212.

eriktrom:
this should be fixed, i rolled back v1.0.23 -> v1.0.21

Issue #84, opened 2019-08-17
the version 1.0.22 do not search from the basePort


AllenField:
the version 1.0.22 do not search from the basePort

BarthesSimpson:
It does, but instead of starting at `basePort p` and checking `p`, `p+1`, `p+2`... `stopPort` in order, it now checks a random sequence of ports in the range `basePort`...`stopPort`. This change was to fix an issue where calling `getPort` more than once in quick succession could lead to the same port being returned multiple times if the client had not finished binding by the time the second call ran its check (https://github.com/http-party/node-portfinder/issues/79).

bcfurtado:
This is a breaking change then since the previously expected behavior was that the `basePort` would be used as the initial default value.
Packages such as [`vue-cli`](https://github.com/vuejs/vue-cli/blob/348e36eedbf77337e372cb7c226509114a1d6f5b/packages/%40vue/cli-service/lib/commands/serve.js#L104) and [`webpack-dev`] were relying on this behavior.

```
$ npm list | grep portfinder
│ ├─┬ portfinder@1.0.22
│ │ ├── portfinder@1.0.22 deduped

$ node --experimental-repl-await
> const portfinder = require('portfinder')
undefined
> portfinder.basePort
8000
> let port = await portfinder.getPortPromise()
undefined
> port
10211
```

```
$ npm list | grep portfinder
│ ├── portfinder@1.0.21 deduped
│ │ ├── portfinder@1.0.21 deduped
├─┬ portfinder@1.0.21

$ node --experimental-repl-await
> const portfinder = require('portfinder')
undefined
> portfinder.basePort
8000
> let port = await portfinder.getPortPromise()
undefined
> port
8000
```

jaredjj3:
> It does, but instead of starting at `basePort p` and checking `p`, `p+1`, `p+2`... `stopPort` in order, it now checks a random sequence of ports in the range `basePort`...`stopPort`. This change was to fix an issue where calling `getPort` more than once in quick succession could lead to the same port being returned multiple times if the client had not finished binding by the time the second call ran its check (#79).

The change made does not solve the problem, since multiple calls of `getPort` can still return the same port. This "port collision" event is more probable as the range `basePort`...`stopPort` decreases to 1.

I recommend reverting this change or making it optional.  The caller will then be responsible for port reservation.

AndreasCag:
I recently configured webpack and was really frustrated with that behaviour.

[...]

PR #86, opened 2019-08-17
Avoid breaking existing consumers who use startPort >= 40000

BarthesSimpson:
This would fix the issue identified in https://github.com/http-party/node-portfinder/issues/85. If there is an existing consumer who uses `getPort` with a `startPort` greater than 40000, instead of throwing an Error this will log a warning to the console and bump the `stopPort` up to 65535 (i.e. behave the same as before `exports.highestPort` was reduced to 40000). 

An error is still thrown if the `stopPort` is lower than the `stopPort` when both of these values are below 40000.

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -392,7 +392,9 @@ function randomShuffle(array) {
 function randomShuffle(array) {
   for (var i = array.length - 1; i > 0; i--) {
     var j = Math.floor(Math.random() * (i + 1));
-    [array[i], array[j]] = [array[j], array[i]];
+    var tmp = array[i]
+    array[i] = array[j]
+    array[j] = tmp
   }
 }
 
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -141,14 +141,14 @@ exports.getPort = function (options, callback) {
   }
 
   if (options.startPort > exports.highestPort) {
-    console.warn('Provided options.startPort(' + options.startPort + ') is higher than the recommended stopPort (' + exports.stopPort + '). This may cause issues on some systems');
+    console.warn('Provided options.startPort(' + options.startPort + ') is higher than the recommended stopPort (' + exports.highestPort + '). This may cause issues on some systems');
     if(options.stopPort <= options.startPort) {
       options.stopPort = _highestPort
     }
   }
 
   if(options.stopPort < options.startPort) {
-    throw Error('Provided options.stopPort(' + options.stopPort + 'is less than options.startPort (' + options.startPort + ')');
+    throw Error('Provided options.stopPort (' + options.stopPort + ') is less than options.startPort (' + options.startPort + ')');
   }
 
   if (options.host) {
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -123,7 +123,6 @@ exports.getPort = function (options, callback) {
   options.port   = Number(options.port) || Number(exports.basePort);
   options.host   = options.host || null;
 
-
   if (!options.stopPort) {
     options.stopPort = Number(exports.highestPort);
   } else {
@@ -139,6 +138,13 @@ exports.getPort = function (options, callback) {
       throw Error('Provided options.stopPort(' + options.stopPort + 'is less than options.startPort (' + options.startPort + ')');
     }
   }
+
+  if (!options.stopPort) {
+    options.stopPort = Number(exports.highestPort);
+  } else {
+    options.stopPort = Math.min(options.stopPort, Number(exports.highestPort))
+  }
+
 
   if (options.host) {
 
@@ -210,6 +216,8 @@ exports.getPort = function (options, callback) {
 
     debugGetPort("in eachSeries() result callback: openPorts is", openPorts);
 
+
+    
     if (openPorts[0] === openPorts[openPorts.length-1]) {
       // if first === last, we found an open port
       if(openPorts[0] <= options.stopPort) {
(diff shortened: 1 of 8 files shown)


BarthesSimpson:
Updated to also address https://github.com/http-party/node-portfinder/issues/84. By default, `getPort` will use the previous sequential scan. If `useRandom: true` is passed inside the options, then the random search will be used instead.

jaredjj3:
LGTM

Thank you for working on this PR.

lib/portfinder.js
```diff
@@ -135,11 +134,23 @@ exports.getPort = function (options, callback) {
...
-    if(options.stopPort < options.startPort) {
-      throw Error('Provided options.stopPort(' + options.stopPort + 'is less than options.startPort (' + options.startPort + ')');
+  }
+
+  if (options.startPort > _highestPort) {
+    throw Error('Provided options.startPort(' + options.startPort + ') is higher than the maximum available port (' + _highestPort + ').');
+  }
+
+  if (options.startPort > exports.highestPort) {
+    console.warn('Provided options.startPort(' + options.startPort + ') is higher than the recommended stopPort (' + exports.stopPort + '). This may cause issues on some systems');
```
jelhan: `exports.stopPort` is `undefined`:

> Provided options.startPort(40001) is higher than the recommended stopPort (undefined). This may cause issues on some systems

It should be `export.highestPort`, isn't it?

lib/portfinder.js
```diff
@@ -135,11 +134,23 @@ exports.getPort = function (options, callback) {
...
+    if(options.stopPort <= options.startPort) {
+      options.stopPort = _highestPort
     }
   }
 
+  if(options.stopPort < options.startPort) {
+    throw Error('Provided options.stopPort(' + options.stopPort + 'is less than options.startPort (' + options.startPort + ')');
```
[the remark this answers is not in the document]
BarthesSimpson: 👍🏼

lib/portfinder.js
BarthesSimpson: Oops - yes, good catch.

jelhan:
I see some strange behavior for `port = 40000`. I'm not 100%ly sure yet if this is a bug or just coincidence. If calling `ember serve --port 40000`, which should equal `getPort({ port: 40000 })` I'm seeing `No open ports found in between 40000 and 40000`. It's working fine for `39999` and `40001`. Haven't have the time yet to follow the investigate the control flow for that edge case yet.

BarthesSimpson:
I think that's because the `startPort:stopPort` interval is closed, so if `startPort` and `stopPort` are the same, no port will be checked. Let me double check that the interval is supposed to be open and then I can update this PR accordingly (later today).

BarthesSimpson:
OK, yeah the interval should be open, not closed. Will update now.

lib/portfinder.js
```diff
@@ -157,10 +168,15 @@ exports.getPort = function (options, callback) {
...
-  for (i = options.port; i < options.stopPort; i++) {
+  for (i = options.port; i <= options.stopPort; i++) {
     candidatePorts.push(i)
   }
-  randomShuffle(candidatePorts)
+
+  if (options.useRandom) {
+    randomShuffle(candidatePorts)
+  } else {
+    candidatePorts.reverse()
```
Ten0: Why is the array reversed? Are they not tested in order afterwards? Were they not tested in order before?

sodatea:
Ping @eriktrom. Sorry to bother you, but is there any chance to get this PR merged soon? Lots of downstream repositories are relying on this feature to work correctly.

Ten0:
> Ping @eriktrom. Sorry to bother you, but is there any chance to get this PR merged soon? Lots of downstream repositories are relying on this feature to work correctly.

+1. With 877 dependant packages and 4 685 655 weekly downloads, this as-patch breaking change that prevents port choice is breaking so many systems right now and shouldn't stay broken for so long.

soloitguy:
Downgrading to 1.0.21 explicitly is a temporary fix.  Our outage was a bit too long to wait for the PR.

eriktrom:
@BarthesSimpson - thanks - i will take a closer look tonight, tomorrow and follow up.

For today, i rolled back to v1.0.21 (as v1.0.23)

eriktrom:
@BarthesSimpson - I feel like we should consider an option that does not directly modify the main repo -- maybe another export `getRandomPort` which uses the in place randomization (port 0 actually may do this automatically, it does in ember-cli, it's been years but I believe its a straight passthrough to portfinder `port: 0` or `port: undefined`. 

I'm very hesitant to release another version that is not a semver major bump unless we can guarantee that changes are strictly additive. 

Thoughts?

eriktrom:
>  If useRandom: true is passed inside the options, then the random search will be used instead.

The only issue here is that, for example, lets say you start 2 test servers inside ember cli - if we pass `useRandom: true` then we have detect the os arch to be darwin - which seems reasonable - unless via docker or minikube your actually using hyperkit - with port forwarding - to be more clear on the port 40,000-60,000 thing - the error is a permission error - binding to the socket works, using it throws a permission error

b/c of matrix of use cases upstream are very high, this single example should help explain my concern. 

fixing for ember-cli is one thing, for firebase, circleci and other tools that depend on the current behavior 'as is' -- i think we make a new method or (if the code was cleaner) an extension poiint, but new method more likely in this case

BarthesSimpson:
Sure - I'll take another look this weekend and see if I can find a good approach for adding a `getRandomPort` export.

eriktrom:
@BarthesSimpson - see #91 for thoughts on releasing changes to portfinder moving forward - mostly that we should only do so if we do a semver major release, reducing exposure to *how* this library has been used by different consumers over the past many years. 

If we go that route, we can consider adding deprecation warning to v1.x that would not take effect until v3.x - although someone would need to take the lead on this as I'm involved in the depths of a few other topics right now - just fyi - open a meta ticket if you want to start a discussion outside this thread regarding such things

I'm going to close this issue and link to it from a clean issue, so we don't spam

eriktrom:
@BarthesSimpson - #92 is a meta issue on a semver major that I would feel comfortable releasing these changes to. Re-opening, with 2.x tag.

eriktrom:
FYI: if interest remains in merging this (w/ the randomization feature via @BarthesSimpson original work) see #105 and vote. I'm taking a quick pulse of the traction it would garner if made possible to you and others.

eriktrom:
closing to cleanup issue list, comments or questions still welcome, if needed

consolidated via #99

PR #101, opened 2020-04-10
mkdirp: 0.5.5

msbit:
Update `mkdirp` to 0.5.5 which in turn updates `minimist` to 1.2.5
which addresses:

    https://www.npmjs.com/advisories/1179

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -105,16 +105,16 @@
       }
     },
     "minimist": {
-      "version": "0.0.8",
-      "resolved": "https://registry.npmjs.org/minimist/-/minimist-0.0.8.tgz",
-      "integrity": "sha1-hX/Kv8M5fSYluCKCYuhqp6ARsF0="
+      "version": "1.2.5",
+      "resolved": "https://registry.npmjs.org/minimist/-/minimist-1.2.5.tgz",
+      "integrity": "sha512-FM9nNUYrRBAELZQT3xeZQ7fmMOBg6nWNmJKTcgsJeaLstP/UODVpGsr5OhXhhXg6f+qtJ8uiZ+PUxkDWcgIXLw=="
     },
     "mkdirp": {
-      "version": "0.5.1",
-      "resolved": "https://registry.npmjs.org/mkdirp/-/mkdirp-0.5.1.tgz",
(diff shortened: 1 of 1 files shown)


eriktrom:
@msbit - gratzi!

LostInBrittany:
Why are you only bumping-up the `package-lock.json`? As `package.json` still depends on `mkdirp` 0.5.1, when we get `node-portfinder` as a dependency, we still get an old version of `minimalist`. 

Would you mind if I did a PR for that?

msbit:
https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004

LostInBrittany:
The dependency is currently `^0.5.1`, and that version has a vulnerability problem, so I don't see why we shouldn't bump it up to `^0.5.5`. If not, why it is at `^0.5.1` and not `^0.5.0`?

LostInBrittany:
BTW, I came here with this request because in several of my projects using `es-dev-server`, I got a badd `minimist` version, and `npm ls minimist` told me the source was the `mkdirp` version we had here...

msbit:
> Looking at the package-lock.json, mkdirp is resolved to 0.5.1 and not 0.5.5 as I should have expected...

I would have expect that too, given the rules for version strings with carets.

From a blank slate:

```
$ mkdir node-portfinder-mkdirp-test
$ pushd node-portfinder-mkdirp-test
$ npm init -y
$ npm add es-dev-server
$ npm ls mkdirp
node-portfinder-mkdirp-test@1.0.0 /Users/tom/Development/third-party/node-portfinder-mkdirp-test
└─┬ es-dev-server@1.56.1
  └─┬ portfinder@1.0.27
    └── mkdirp@0.5.5
```

Could it be that you had `mkdirp` already installed before updating `portfinder`?

Anyhow, looks like you've got a PR there to get the `package.json` bumped too, which should address your issue.

LostInBrittany:
I don't understand why, I simply did a `npm update`  on two projects using `es-dev-server` and `es-dev-server` itself got updated to 1.56.1, `portfinder` to 1.0.27 but `mkdirp` got stuck on 0.5.1.  

As I got that twice, I suppose that other people could have the same problem, sur bumping-up `package.json` seemed the safe thing to do. But I am still a bit puzzled about why...

eriktrom:
FYI - thanks for discussing this

PR #103, opened 2020-05-14
Add missing highestPort property to TypeScript definition

brunogaspar:
Was causing warning about missing property.

#### Before:

before

#### After:

after

Signed-off-by: Bruno Gaspar <brunofgaspar1@gmail.com>

### DIFF
--- a/lib/portfinder.d.ts
+++ b/lib/portfinder.d.ts
@@ -34,6 +34,11 @@ interface PortFinderOptions{
 export let basePort: number;
 
 /**
+ * The highest port to end any port search from.
+ */
+export let highestPort: number;
+
+/**
  * Responds with a unbound port on the current machine.
  */
 export function getPort(callback: PortfinderCallback): void;


eriktrom:
Ideally we want to make sure that the highest port does not exceed 65536 as this is the highest number of open file descriptors most development systems can have open. (an int in C is 4 bytes * (8 bits in a byte) == 4**8 = 65536

there is also a limitation with macos as it reserves the top 20,000 ports or so for root (wheel) access only. It uses these for the touchbar, which is controlled via a protocol that uses an FD as well.

that said, i don't recall off hand if typescript can type alias number with restrictions on the range or max size of a number, but if so, that'd be ideal... 

(I'll check it out in the next 1.5 weeks, if you know, drop me a link or note)


thanks for the contribution @brunogaspar

brunogaspar:
Hi @eriktrom 

That does not seem entirely possible at the moment (it's quite limited), at least according to this open issue https://github.com/Microsoft/TypeScript/issues/15480

akkarapon26 (approved):

Issue #104, opened 2020-05-19
Free port still in use


sla89:
Hello together,

I am facing the problem that portfinder is blocking a free port because the callback is called directly after calling `server.close()`. 

The problem is that  `server.close()` is an asynchronous operation and as long as the  `close` event is not emitted the port is still used by the server (Node API docs: https://nodejs.org/docs/latest/api/net.html#net_server_close_callback):

```
server.close([callback])[src]#
Added in: v0.1.90
callback <Function> Called when the server is closed
Returns: <net.Server>
Stops the server from accepting new connections and keeps existing connections.
This function is asynchronous, the server is finally closed when all connections are
ended and the server emits a 'close' event. The optional callback will be called once
the 'close' event occurs. Unlike that event, it will be called with an Error as its only
argument if the server was not open when it was closed.
```

This is where the problem happens:
https://github.com/http-party/node-portfinder/blob/master/lib/portfinder.js#L42

With regards,
Stefan

rathboma:
I am also having this issue as I am immediately starting a server to provide SSH tunnel access, and every now and then it fails due to the port not being available.

eriktrom:
> The problem is that server.close() is an asynchronous operation and as long as the close event is not emitted the port is still used by the server (Node API docs:

> I am also having this issue as I am immediately starting a server to provide SSH tunnel access, and every now and then it fails due to the port not being available.

When using async i/o, which was introduced along with OS scheduling (by different names, depending on the OS) -- one of the downsides is that it is up to the operating system to handle such things. The os will ask the kernel for permission to both open and close a file descriptor, and upon giving permission to open a file descriptor, the os will schedule to do so with a priority. The same goes for free'ing a file descriptor, of which http/sockets are bound to. 

thus, tl;dr - besides being strongly and thus correctly ordered (on x86) - there is no way to know exactly when the release of the file descriptor will occur, be cleaned up, and known to the kernel to be available for use again.

that's the tl;dr - the longer story is much more complicated.

my suggestion is to poll and use a Promise wrapping setTimeout/setInterval. You can wrap the interval in a promise, when it rejects, recurse, when it succeeds, clear the interval, use the port.

Let me know if you have further questions, sorry for the delay here!

eriktrom:
closing to cleanup issue list, comments or questions still welcome, if needed

Issue #99, opened 2019-10-24
Before filing a socket not closed issue: understanding async I/O + ordering + OS differences + stability requirements


eriktrom:
write contributor guidelines, including emphasis on development of port finding:

- [ ] stability over speed, 
- [ ] stability over features that don't serve the larger portion of the consumer base (aka, wrap this lib if you want to extend it, provide examples and how to's for those in need).
- [ ] edge cases (per OS, mentioned many times in discussion, but mainly 1) macOS touch bar takes ports 40,000-60,000, so exclude those) 2) Windows (in the past) has allowed port rebinding
- [ ] How to wrap portfinder for your own custom need  (in order to leave this lib encapsulated to one task that is done well (unix/substack philosophy).

---- 
>
Summary: (from https://github.com/http-party/node-portfinder/pull/105#issuecomment-642807837)
>
> Here is what I would consider:
>
> not modifying this library, except under one circumstance
that another library, call it portlocker™️, that wraps port finder to find a port, but then also maintains that open socket, which can then be used via a callback
--
Reasoning:
>
> We have no idea how this lib is being consumed - with 7+ million npm downloads a week, we don't want to break anything nor anyones expectations of its current behavior
>
> Such a library, like portlocker™️, can then do one thing and do it well - handle the happens before ordering of async i/o scheduling. This would allow risk free iteration on current and future needs, anything from single threaded async i/o to multi (worker or process) async i/o could be explored. There has been some interest in this by others, in various forms, none of which are core to portfinder's original goals or intents, but which could easily consume portfinder and add such things.
>
> If this proposal gains traction, I will consider. Provide a 👍, and I'll tally the priority and interest.>
>
> Keeps the comments light, but feel free to weigh in

eriktrom:
Consolidate with:

- [ ] all cross refs found at #105 that are 
- [ ] move ad hoc summary and proposal found randomly in comment @ #105: https://github.com/http-party/node-portfinder/pull/105#issuecomment-642807837

eriktrom:
too much scope, closing for now

PR #105, opened 2020-05-19
Wait until server is closed before invoking callback.

sla89:
Otherwise the port may be blocked because it is signaled to be
free while server was not yet closed.

See bug #104

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -38,9 +38,11 @@ internals.testPort = function(options, callback) {
   function onListen () {
     debugTestPort("done w/ testPort(): OK", options.host, "port", options.port);
 
-        options.server.removeListener('error', onError);
-        options.server.close();
+    options.server.removeListener('error', onError);
+    options.server.close(function () {
+      debugTestPort("done w/ testPort(): Server closed", options.host, "port", options.port);
       callback(null, options.port);
+    });
   }
 
   function onError (err) {


rathboma:
This seems like a good change. I've seen this error materialize several times -- we test for an open port, then extablish a server to do ssh tunneling. There seem to be ephemeral errors, and this so I guess this is why

lib/portfinder.js
```diff
@@ -38,9 +38,11 @@ internals.testPort = function(options, callback) {
   function onListen () {
     debugTestPort("done w/ testPort(): OK", options.host, "port", options.port);
 
-        options.server.removeListener('error', onError);
-        options.server.close();
+    options.server.removeListener('error', onError);
+    options.server.close(function () {
+      debugTestPort("done w/ testPort(): Server closed", options.host, "port", options.port);
       callback(null, options.port);
```
eriktrom, line 44: just b/c the we've received a close event doesn't mean resources are fully de-allocated by the OS, it just means that node has notified the os to close, and the os has notified back to node that "go ahead and do things that don't depend on having an open file descriptor" <-- of which does not include "opening another one immediately" unless we wrap this in a try/catch or similar, and recurse - the OS may too be starved. 

in a perfect world, or with sequential consistent ordering (which is slow and x86 makes less needed) this would work.... 

I may have to verify this claim, but in general, with CSP (strongly ordered) scheduling in place, all that matters is that c comes after a and b comes after a but whether c comes after b, is often irrelevant for ordering that of the scheduler, it only cares about the eventual ordering... 

I should probably make sure my brain is not misleading me, in meantime, feel free to comment thoughts, i'll come back next week when time permits (no more than 8 days fyi)

eriktrom: also, if this port really is blocked, running 2 instances of in 2 shells lets say, would yield indeterminate results as to which terminal wins, again due to os scheduling (which I suppose is to say, if you wrote a test with 2 child processes (say process x and y) trying to aquire the lock on the FD, 100 times in a row, the results, would be 50/50 as to which child process wins and gets to use port 22(in this example) - which is just another explanation for how os scheduling that is not strongly ordered (which is expensive hence the x86 optimizations to auto schedule this, without breaking eventual consistency) of the total order. 

I know its a bit confusing, this feature dropped in the late 90's IIRC... 

I will look deeper though, but wanted to add another example, i may be forgetting something that makes this easier, if so, jog my memory in the meantime :)

eriktrom:
EDIT: (previous comments we're not clear in the direction I would consider taking this, I have replaced them below).


Here is what I would consider:

- not modifying this library, except under one circumstance
   - that another library, call it `portlocker`™️, that wraps port finder to find a port, but then also maintains that open socket, which can then be used via a callback

--
Reasoning:

1) We have no idea how this lib is being consumed - with 7+ million npm downloads a week, we don't want to break anything nor anyones expectations of its current behavior

2) Such a library, like `portlocker`™️, can then do one thing and do it well - handle the happens before ordering of async i/o scheduling. This would allow risk free iteration on current and future needs, anything from single threaded async i/o to multi (worker or process) async i/o could be explored. There has been some interest in this by others, in various forms, none of which are core to portfinder's original goals or intents, but which could easily consume portfinder and add such things.

If this proposal gains traction, I will consider. Provide a 👍, and I'll tally the priority and interest.

Keeps the comments light, but feel free to weigh in

eriktrom:
closing to cleanup issue list, comments or questions still welcome, if needed

Consolidated via #99

PR #114, opened 2021-01-31
Bump debug to latest

lopopolo:
If this patch is merged and accepted, could the maintainers push a release to npm?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,11 +35,11 @@
       "dev": true
     },
     "debug": {
-      "version": "3.2.6",
-      "resolved": "https://registry.npmjs.org/debug/-/debug-3.2.6.tgz",
-      "integrity": "sha1-6D0X3hbYp++3cX7b5fsQE17uYps=",
+      "version": "4.3.1",
+      "resolved": "https://registry.npmjs.org/debug/-/debug-4.3.1.tgz",
+      "integrity": "sha512-doEwdvm4PCeK4K3RQN2ZC2BYUBaxwLARCqZmMjtF8a51J2Rb0xpVloFRnCODwqjpwnAoao4pelN8l3RJdv3gRQ==",
       "requires": {
-        "ms": "^2.1.1"
+        "ms": "2.1.2"
       }
     },
     "diff": {
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   ],
   "dependencies": {
     "async": "^2.6.2",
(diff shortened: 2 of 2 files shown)


lopopolo:
The build fails for Node.js 0.12, 4.x, and 5.x, all of which are EOL.

eriktrom:
> The build fails for Node.js 0.12, 4.x, and 5.x, all of which are EOL.

True, but we support node v0.10 - current. This project is one of the lowest deps in the tree of any project, if the whole world has not moved on to node lts, then portfinder can't move to lts only. In fact it does not need to.

What I can do however is make sure that the lockfile for portfinder is correct, meaning portfinder will look for the debug module under ./node_modules/portfinder/node_modules/debug. In your own project, make sure to add debug (the version of your choice) to your package.json. When done correctly, your project will require('debug'), when doing so it will look up debug from ./node_modules/debug/index.js, while portfinder will use its own lookup path, which is always the closest node_modules directory relative to the module doing the require.

Tl;dr - it looks like you got it working by fixing your own project's lockfile, which then installed intermediate node_modules directories for deps of deps like portfinder's use of debug. 

I am most curious about the pain you experienced - do u mind summarizing your experience. If you do, I will repo and then consider options to make consuming this library easier when their are multiple versions of another library (like debug) being used in upstream projects.

Thanks 

In the meantime, in your own project you can use a newer version of

eriktrom:
@lopopolo - your package-lock.json looks correct - https://github.com/artichoke/www.artichokeruby.org/blob/9ed196e8e08f95c6e0213849fc5dd07ce858f607/package-lock.json#L3543

also - there are 9 copies of debug in your project - most of them are version 2.x, a couple are version 3.x, only the top debug library is version 4.x - the one you are using in the project directly - and the older versions are correctly scoped under their respective dependencies.

For example, express.js depends on debug 2.x - in all 7 other dependencies would need to update to version 4.x to match your expectation for how modules are resolved. 

What specifically did you run into with portfinder that you did not run into with express or the other 7 libraries that also use 2.x or 3.x version of debug?

Thanks, any explanation would help clarify why portfinder caused u pain while libraries doing the exact same thing in your package.json file did not.

lopopolo:
I consider duplicate dependencies anywhere in the tree undesirable. That `node_modules` allows it doesn't impact my goal of removing duplicate dependencies.

Thanks for taking a look. If you're not planning on accepting this patch, feel free to close this issue.

PR #115, opened 2021-02-08
Add mutator methods to allow ports ranges to be set with module-style imports

inlined:
Ran into a compiler error when using portfinder in TypeScript. Module exports are considered immutable in typescript, so the following code won't compile:

```
import * as portfinder from "portfinder";

portfinder.basePort = 3000;
```

Instead, a new accessor method allows us to say `portfinder.setBasePort(3000);`

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,8 +48,8 @@ By default `portfinder` will start searching from `8000` and scan until maximum 
 You can change this globally by setting:
 
 ```js
-portfinder.basePort = 3000;    // default: 8000
-portfinder.highestPort = 3333; // default: 65535
+portfinder.setBasePort(3000);    // default: 8000
+portfinder.setHighestPort(3333); // default: 65535
 ```
 
 or by passing optional options object on each invocation:
--- a/lib/portfinder.d.ts
+++ b/lib/portfinder.d.ts
@@ -34,9 +34,19 @@ interface PortFinderOptions{
 export let basePort: number;
 
 /**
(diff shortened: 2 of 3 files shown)


inlined:
Ping?

eriktrom:
yep sorry for the delay

so unfortunately this lib has to work with node going back as far as v0.10.x - as enterprise code bases, believe it or not, sometimes never upgrade node itself, and this lib is used by more packages than one would at first imagine. 

thanks though for being awesome and contributing. despite having to close this, i highly respect those who contribute to open source and I wish I could merge this for that sake alone but unfortunately i can't, there should be a contribution attempt chart b/c i think its such an important part of today's economy and society but until then, thanks will have to do :)

cheers

inlined:
I'm confused. Why does offering a mutator method affect the version of node that can be used? The opposite is true; without this change portfinder won't be able to be used in a modern codebase but with this change it can be used in old or new codebases.

eriktrom:
> I'm confused. Why does offering a mutator method affect the version of node that can be used? The opposite is true; without this change portfinder won't be able to be used in a modern codebase but with this change it can be used in old or new codebases.

Correct. 

PS - Sorry I haven't been around for the last 12 months. Knowing that was going to be the case, I left an issue at https://github.com/http-party/node-portfinder/issues/122 - which until 4 days ago was mostly crickets. I have returned and will address the outstanding PR's and issues. 

Thanks for your patience.

PR #117, opened 2021-05-28
Only import the needed functions from async

aminya:
This reduces the size of portfinder from 34 KB to 14 KB

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -11,9 +11,11 @@ var fs = require('fs'),
     os = require('os'),
     net = require('net'),
     path = require('path'),
-    _async = require('async'),
     debug = require('debug'),
     mkdirp = require('mkdirp').mkdirp;
+
+var eachSeries = require('async/eachSeries');
+var timesSeries = require('async/timesSeries');
 
 var debugTestPort = debug('portfinder:testPort'),
     debugGetPort = debug('portfinder:getPort'),
@@ -142,7 +144,7 @@ exports.getPort = function (options, callback) {
   }
 
   var openPorts = [], currentHost;
-  return _async.eachSeries(exports._defaultHosts, function(host, next) {
+  return eachSeries(exports._defaultHosts, function(host, next) {
     debugGetPort("in eachSeries() iteration callback: host is", host);
 
     return internals.testPort({ host: host, port: options.port }, function(err, port) {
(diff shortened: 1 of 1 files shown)


eriktrom:
sorry for the late reply - i want to make this better, i appreciate the pr - which is completely legit but doesn't actually do anything b/c this is only for development use and even a 0.1% chance risk for a 0% gain b/c this is only for dev means its not worth the risk tradeoff, small as it literally is. 

The codebase is essentially frozen except for new operating systems that might throw new errors i may need to catch, otherwise, there is not any benefit to changing it - unless your using vscode to develop inside a cloud container from a laptop that has a bluetooth tether the first time you install a package that uses it - a very not likely scenario.

I thus consider things like reducing kb's over the wire for this development only portfinder "cosmetic" - it makes it look better when u spit out a pretty dep chart with kb's for each dep, but in reality it just adds risk of a typo or misread by me or the contributor 

(if i merged these everytime they came in, eventually a mistake would be made, as were all human, and the effects would cost companies and orgs real time or money or both so I have to say no but i really do appreciate the contribution

ps - don't use this on ur server in production - it was never intended for that, I know some people are trying to use it for that and you thus could say another reason I don't merge such enhancements is to disway people from using in production or on a live server.  Even in a country with very poor internet, the one time download of this package is far less than the editor plugin used to highlight and theme any editor. 

perspective is hard to explain, hopefully that provides a bit of it though from my point of view.

again, thanks for being awesome and trying though. that's worthy in and of itself, for what its worth. :)

aminya:
Thanks for the response. Given your comment, I will make sure to remove the dependencies on `portfinder` in the packages that I maintain.

eriktrom:
> You should add this information to the Readme to inform others about the locked codebase and its intention.

your right - i inherited it and never documented what I learned about it over the years. that said, in order to use this on a production server you'd have to open the firewall internally anyway and that I feel is enough to give people the hint that they should think twice - for those who know what their doing though - i mean hey, you could use for nefarious purposes or legit purposes from a server... to do good or evil things and thus, it is a legit portfinder.

truth is, how to say that its a low level lib - be creative but also careful with it - i kind of decided if u can read the codebase then u can get as creative with it as u want. If u can't then its just there when you start your react app or submit a pr that circle ci picks up.

you have made the most convincing argument though about this so far... im thinking .. :)

PR #132, opened 2022-08-02
Bump version of async in package.json

MasterOdin:
PR bumps the version of `async` dependency in package.json so that it's not possible to get a vulnerable version of `async` via this dependency. #130 only bumped the version in the `package-lock.json` file which only affects this repository, and not downstream consumers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "lib"
   ],
   "dependencies": {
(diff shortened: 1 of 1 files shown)


eriktrom:
@MasterOdin - gratzi for saving me some time here :)

PR #134, opened 2022-08-02
Modernize - DO NOT MERGE

eriktrom:
Modernization work to allow this package to remain relevant and secure.

This commit is to test the workflow just added.

----

Before merge: 

- [x] Support node lts - check how macOS m1 user experience will be when going below node 12.
            - i personally am waiting for xcode (full version, as xcode-select said I needed it to build node < 12.x)

PR #135, opened 2022-08-02
Add test around recursively making socket directory

MasterOdin:
PR adds a test case that validates the recursive mkdir behavior that currently exists within the `getSocket` function. This helps demonstrate the breaking change in #134 which removes `mkdirp` in favor of the builtin `fs.mkdir` where for node < 10.12, the `recursive` flag didn't exist, and so it would not behave the same as `mkdirp`. To demonstrate that, I merged this code onto the `modernize` branch in my fork to trigger CI, and where all node < 10.x fail: https://github.com/MasterOdin/node-portfinder/actions/runs/2782492617.

### DIFF
--- a/test/port-finder-socket-test.js
+++ b/test/port-finder-socket-test.js
@@ -69,7 +69,15 @@ function stopServers(callback, index) {
 }
 
 function cleanup(callback) {
-  fs.rmdirSync(badDir);
+  if (fs.existsSync(path.join(badDir, 'deeply', 'nested'))) {
+    fs.rmdirSync(path.join(badDir, 'deeply', 'nested'));
+  }
+  if (fs.existsSync(path.join(badDir, 'deeply'))) {
+    fs.rmdirSync(path.join(badDir, 'deeply'));
+  }
+  if (fs.existsSync(badDir)) {
+    fs.rmdirSync(badDir);
+  }
   stopServers(callback, 0);
 }
 
@@ -83,7 +91,7 @@ vows.describe('portfinder').addBatch({
           }, this.callback);
         }.bind(this));
       },
-      "the getPort() method": {
+      "the getSocket() method": {
         topic: function () {
           portfinder.getSocket({
             path: path.join(socketDir, 'test.sock')
@@ -100,7 +108,7 @@ vows.describe('portfinder').addBatch({
   "When using portfinder module": {
     "with no existing servers": {
       "the getSocket() method": {
-        "with a directory that doesnt exist": {
+        "with a directory that doesn't exist": {
           topic: function () {
             fs.rmdir(badDir, function () {
               portfinder.getSocket({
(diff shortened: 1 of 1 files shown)


test/port-finder-socket-test.js
```diff
@@ -69,7 +69,15 @@ function stopServers(callback, index) {
...
-  fs.rmdirSync(badDir);
+  if (fs.existsSync(path.join(badDir, 'deeply', 'nested'))) {
+    fs.rmdirSync(path.join(badDir, 'deeply', 'nested'));
+  }
+  if (fs.existsSync(path.join(badDir, 'deeply'))) {
+    fs.rmdirSync(path.join(badDir, 'deeply'));
+  }
+  if (fs.existsSync(badDir)) {
+    fs.rmdirSync(badDir);
+  }
```
MasterOdin, line 80: Without adding the `if (fs.existsSync(...))` checks, when my new test failed on `modernize`, the `fs.rmdirSync` call would throw as the nested directory doesn't exist, and so then `stopServers` wouldn't be called and then vows would hang as there'd be sockets that were never cleaned up.

MasterOdin: Unfortunately, that function was added in Node 5, so until 0.12 - 4.x are no longer tested, I couldn't incorporate that. While I do think could probably revisit it later once that does happen, good to expand the test suite to help ensure no breaking changes on modernizing (well, besides dropping the old versions).

eriktrom: @MasterOdin - so the issue where the left over bad folders and sockets inside fixtures occurred again - only once, this time on a cold mac m1. Recall though, last night it happened (consistently) inside docker for mac, intel x86

I'm noting it, I'll revisit, I may know why. 

Thoughts welcome. Forgive the short note, snooze time.

eriktrom: it also passed completely for me earlier. 

maybe i botched it, i'll check tomorrow. 

feel free to check if ur curiuos

eriktrom: @MasterOdin 

unfortunately the recursive deletion code didn't work - i experienced a hang and left over dirty file system in the test directory 

what u think about using  [fs.mkdtemp(prefix[, options], callback)](https://nodejs.org/api/fs.html#fsmkdtempprefix-options-callback) as a lifeline here? If people are using node < 5, they better have their own npm server :)  

its the best option for the tradeoff curve

(ps - to reproduce what i see, try in docker, rebuild each time you run, you'll notice - its not consistent below node 12 ish)

eriktrom, line 80: @MasterOdin - here (https://github.com/http-party/node-portfinder/issues/163)

test/port-finder-socket-test.js
```diff
@@ -113,6 +121,24 @@ vows.describe('portfinder').addBatch({
             assert.equal(socket, path.join(badDir, 'test.sock'));
           }
         },
+        "with a nested directory that doesn't exist": {
```
eriktrom, line 124: @MasterOdin - here (https://github.com/http-party/node-portfinder/issues/163)

PR #137, opened 2022-08-06
Modernize2

eriktrom:
supersedes #134 
closes #134 

This PR contains a few simple commits meant to modernize the dependencies that had security vulnerabilities in them.

It also contains some code from @MasterOdin that attempts to fix the test setup. Note that I have left (but upgraded) mkdirp for now. We can revisit that a week from now. I will not be in town next week, so this release needs to be minimal as to not interrupt my vacation :). 

With only dependency updates included in the main code, things should not change behavior wise in any shape or form. This should be completely safe and non breaking to merge. I will do so now, and make a release.

cc @MasterOdin fyi

ps tests all pass :) - we probably don't need the recursive fs.rmdir code but let me know your thoughts - it's only for the tests and I need to get these security updates out - I'll be around next weekend(a week from tomorrow) - I'll look deeper then, thanks for all your support and help lately ;)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [0.12.x, 4.x, 14.x, 16.x, 18.x]
+        node-version: [0.12.x, 4.x, 6.x, 8.x, 10.x, 12.x, 14.x, 16.x, 18.x]
         os: [ubuntu-latest, macos-latest, windows-latest]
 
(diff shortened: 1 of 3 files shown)


eriktrom:
Released to npm as v1.0.29

MasterOdin:
In my opinion, you'd ideally have some code to clean up the folders that are made during the test, be it if you're making them in a known spot or using mktempdir function. However, if you're willing to make minimum support to 12.x (which is the oldest currently supported LTS), could use the `recursive` option to `fs.rmdir`.

PR #139, opened 2022-08-07
fix(types): fixed typescript definition

jonluca:
The typescript definition was missing the `function` keyword.

### DIFF
--- a/lib/portfinder.d.ts
+++ b/lib/portfinder.d.ts
@@ -6,7 +6,7 @@
 
 type PortfinderCallback = (err: Error, port: number) => void;
 
-interface PortFinderOptions{
+interface PortFinderOptions {
   /**
    * Host to find available port on.
    */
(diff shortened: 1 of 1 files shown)


lib/portfinder.d.ts
```diff
@@ -46,7 +46,7 @@ export let highestPort: number;
 /**
  * Set the higheset port to end any port search from.
  */
-export setHighestPort(port: number);
+export function setHighestPort(port: number);
```
eriktrom, line 49: @jonluca - can we make this not pass CI when this fails like it did?

eriktrom: if u don't have time just let me know, otherwise, gratzi

eriktrom: nvm #141 looks to address that

eriktrom (approved):

PR #140, opened 2022-08-07
Write in typescript and ES2020, create ESM output

jonluca:
This is a PR that's born out of the previous fix - it didn't make sense to ship separate types, that were missing some options and had to be manually updated.

This rewrites the core logic with strongly typed typescript, and modernizes some of the code.

The build outputs both es3 and es2020, so this is fully backwards compatible.


eriktrom:
closing in favor of #139

eriktrom:
wow thanks for the work, but as mentioned in #141 - the simple - which is all we need - type testing in that pr would not work if we completely rewrite everything. 

nice work though, good practice is never a waste of time :)

PR #146, opened 2023-01-08
Add promise version of getPorts function

MasterOdin:
PR adds a new `getPortsPromise` function to allow using a promisified version of `getPorts`, similar to how there's a `getPortPromise` version of `getPort`. A test is included for the function, which mimics what I saw for testing `getPortPromise`.

I also fixed a type error which made it seem like the `options` parameter was required for `getPorts`, where it's optional, just need to always pass the `count` and `callback`.


eriktrom:
This change looks good. Thanks.

PR #148, opened 2023-05-26
Fix adding options.host to defaultHosts list

MasterOdin:
PR fixes a bug reported in https://github.com/http-party/node-portfinder/pull/120#issuecomment-1502633024, where the current `indexOf` check will only succeed if `options.host` was already in `exports._defaultHosts`, as opposed to the desired behavior of appending it if it's _not_ in the array.

I've also cleaned up the nested conditional to just be one level as the nesting was not necessary.

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -141,10 +141,8 @@ exports.getPort = function (options, callback) {
     }
   }
 
-  if (options.host) {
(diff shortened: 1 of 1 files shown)


eriktrom:
Thank you for this bug fix. Good work.

PR #149, opened 2023-07-18
Wait until server is closed before invoking callback

Pwndrian:
With the current implementation, there is a chance that a user binds a socket on a port (that was reported as being available) although the server (created to check whether that port is available) is not fully closed yet. This causes the binding to fail, because the port is still in use.

In this PR, we change the implementation such that the the callback passed to the `testPort` is invoked when the server is fully closed. Thanks to @sla89 for the idea!

Unfortunately, we didn't manage to write a unit tests that reliably verifies the correct behavior. I'm open for ideas ;)


eriktrom:
Yes this is correct, thank you for your contribution.

Issue #153, opened 2025-03-03
Upgrade async to Latest Version to Address CVE-2024-39249


TimothyAsirJeyasing:
Currently, portfinder is using an outdated version of `async`, which is known to have a security vulnerability identified as CVE-2024-39249. This vulnerability poses a risk and should be addressed promptly by upgrading to the latest version of `async`.

**Steps to Reproduce:**
1. Clone the portfinder repository.
2. Check the version of `async` in `package.json`.
3. Note that it is using a version vulnerable to CVE-2024-39249.

**Expected Behavior:**
Portfinder should use the latest version of `async` to ensure security and stability.

**Proposed Solution:**
Upgrade the `async` dependency to the latest version in `package.json` and run `npm install` to update the lock file accordingly.

```json
  "dependencies": {
    "async": "^latest_version"
  }

MasterOdin:
Moving to the async 3.x would require this library dropping support for Node <= 4, per [their changelog](https://github.com/caolan/async/blob/master/CHANGELOG.md):

> We are now making heavy use of ES2015 features, this means we have dropped out-of-the-box support for Node 4 and earlier, and many old versions of browsers. (#1541, #1553)

However, per [this issue](https://github.com/caolan/async/issues/1975), the vulnerability requires really convoluted method to hit the exploit that it's very unlikely to happen in production unless a developer did it to themselves.

MasterOdin:
If I were to argue for doing this, it would be that:
1. Node 4 has been end-of-life for over 6 years and that anyone still maintaining a codebase that requires it probably is not updating their dependencies anymore either, so breaking backwards compatibility for it wouldn't really affect anyone.
2. Seeing a vulnerability warning whenever someone installs this dependency via npm (or using `npm audit`) and that most users only see that there's a vulnerability and don't realize how minor or whatever it actually is, and so you'll end up with these sorts of issues / comments.

If the library were to drop support for Node < 10, then could minimally do the following cleanup as well:
1. Remove the `mkdirp` dependency in favor of using the `recursive` option to `fs.mkdir`
2. Checking for `Promise` exists can be removed

eriktrom:
> Node 4 has been end-of-life for over 6 years and that anyone still maintaining a codebase that requires it probably is not updating their dependencies anymore either, so breaking backwards compatibility for it wouldn't really affect anyone

[...]

PR #156, opened 2025-03-06
move test suite to jest

### LINKED ISSUE #155: Replace vows test runner
`vows` throws an error when trying to run it using Node > 10 and there's a test failure due to this line:

https://github.com/vowsjs/vows/blob/cbd4b9f670bcd0085f311005e5da436354607a68/lib/vows.js#L208

where [`utils.puts`](https://nodejs.org/docs/latest-v10.x/api/util.html#util_util_puts_strings) was deprecated in Node 10 and removed in 12. Given that the codebase has not seen a commit in over 5 years, I think it's probably worth switching to something that's actively supported.

Some popular alternatives that I'm familiar with would be:
* [jest](https://www.npmjs.com/package/jest)
* [vitest](https://www.npmjs.com/package/vitest)
* [mocha](https://www.npmjs.com/package/mocha) (paired with [chai](https://www.npmjs.com/package/chai) usually)

If support for Node < 18 was dropped, could also look to use the built-in [node test runner](https://nodejs.org/docs/latest-v18.x/api/test.html) given the relatively simplistic needs of this library's tests.

MasterOdin:
Closes #155 

PR moves the test suite over from [vows](http://npmjs.com/package/vows) (which is defunct and doesn't work properly on Node >= 10) to [jest](http://npmjs.com/package/jest). I did have to use a somewhat older version of jest to run against older EOL versions of node. Based on https://github.com/http-party/node-portfinder/issues/153#issuecomment-2702254544, I chose the newest version that would still allow for testing and specifying node >= 6 in the engines field (once `async` is updated).

This does make `npm audit` complain about a number of vulnerabilities in installed packages, but given this is just within the test suite, I think that's fine. There's also the downside that adding jest does add quite a bit of dependency tree (whereas `vows`  was relatively small), but I think using vitest or mocha would add similar amount of weight to the repo.

I did also have to update the test files names so that the suffix was `.test.js` instead of `-test.js` which allows jest to automatically pick up the files without having to modify the [`testMatch`](https://jestjs.io/docs/configuration#testmatch-arraystring) setting.


.github/workflows/ci.yml
```diff
@@ -40,6 +36,10 @@ jobs:
         node-version: ${{ matrix.node-version }}
         cache: 'npm'
 
+    - name: Update npm
+      if: matrix.node-version == '6.x'
+      run: npm install -g npm@^5
```
MasterOdin, line 41: node 6.x installs npm@3 which had issues installing all dependencies of jest. Updating to npm@5 fixes the issue, and didn't think it worth spending time trying to figure out why it wasn't working.

.github/workflows/ci.yml
```diff
@@ -65,4 +65,4 @@ jobs:
     - name: Install typescript
       run: npm install --save-dev typescript
 
-    - run: node_modules/.bin/tsc --strict lib/portfinder.d.ts
+    - run: node_modules/.bin/tsc --strict --lib es2015 lib/portfinder.d.ts
```
MasterOdin, line 68: Needed to add `--lib es2015` here as some of the dependencies that jest brings in uses es2015 features (e.g. `Set`, `Map`, etc.) and without this would get an error. Node 6+ is largely es2015 complete, so not really a compatibility risk.

package.json
```diff
@@ -21,15 +21,15 @@
...
   },
   "devDependencies": {
-    "vows": "^0.8.3"
+    "jest": "^24.9.0"
   },
   "main": "./lib/portfinder",
   "types": "./lib/portfinder.d.ts",
   "scripts": {
-    "test": "vows test/*-test.js --spec"
+    "test": "jest --runInBand"
```
MasterOdin, line 29: Jest will try to run the test files in parallel via workers, which breaks the suite given that tests re-use the same ports. Given how quickly it runs (< 1s), I think the lack of parallelization doesn't matter.

test/port-finder-integration.test.js
MasterOdin, line 1: Jest does not allow specifying an order for test files to run, and will run them randomly. In testing, I didn't find that it would matter when this particular test would run though.

eriktrom:
👀 tomorrow am. Thanks again. This should help new people a lot.

eriktrom (approved):
Very good work. Thank you again.

PR #157, opened 2025-03-07
Update async to ^3.2.6

### LINKED ISSUE #153: Upgrade async to Latest Version to Address CVE-2024-39249

MasterOdin:
Closes #153 

PR updates async dependency to latest version. The only change that really affects this library was that the first function argument to [`whilst`](https://caolan.github.io/async/v3/docs.html#whilst) now has a callback function that need to call to return the test result, vs just returning it as before with v2.

Per the discussion in the linked issue, need to drop support for Node <= 4 as async dropped support for those versions of node.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.0.33",
       "license": "MIT",
       "dependencies": {
-        "async": "^2.6.4",
+        "async": "^3.2.6",
         "debug": "^3.2.7",
         "mkdirp": "^0.5.6"
       },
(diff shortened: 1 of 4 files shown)


eriktrom:
Thanks for rebasing against the new jest branch(now in master). This looks good once those changes are in.

MasterOdin:
Done @eriktrom.

eriktrom (approved):
Nice work, thanks @MasterOdin

PR #154, opened 2025-03-06
docs: clean up readme

Electroid:
Hi there, I made some minor readability changes to the readme. 

* Fixed copy/paste of commands from including the "$"
* Removed leading indent from code examples
* Mention yarn and bun

PR #161, opened 2025-03-08
Replace var with const/let

MasterOdin:
PR replaces usage of `var` with `const`/`let` which were introduced in Node 6. I didn't touch the `var` in the README as that was done in #154.


eriktrom:
Very sweet

PR #162, opened 2025-03-08
Update debug to 4.3.6

MasterOdin:
PR updates the library to use `debug:^4.3.6` as dependency. This is the latest version (it's 2 behind), but I chose this version as it's the latest version that fixed JS deprecations or reported vulnerabilities, while still giving a bit of leeway to folks to dedupe if they already have a recentish version of debug installed.

There are no direct tests for debug, so I tested it by doing `DEBUG=portfinder:* npm run test` and confirming that the output remained the same between versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "async": "^3.2.6",
-        "debug": "^3.2.7",
+        "debug": "^4.3.6",
         "mkdirp": "^0.5.6"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


eriktrom:
>  I chose this version as it's the latest version that fixed JS deprecations or reported vulnerabilities, while still giving a bit of leeway to folks to dedupe if they already have a recentish version of debug installed.

I agree 

> There are no direct tests for debug, so I tested it by doing DEBUG=portfinder:* npm run test and confirming that the output remained the same between versions.

Thanks for checking :)

MasterOdin:
@eriktrom rebased

eriktrom:
Gratzi!

PR #165, opened 2025-03-11
Remove mkdirp dependency

### LINKED ISSUE #163: Remove mkdirp, upgrade node to >=10.12.0, use built-in node mkdirp
@MasterOdin - I think we should upgrade node to `10.12.0` and then remove mkdirp. 

We went through this idea already and you and I already wrote this code

https://github.com/http-party/node-portfinder/pull/134/files
https://github.com/http-party/node-portfinder/pull/135/files

MasterOdin:
Closes #163 

PR removes the `mkdirp` dependency relying on `{ recursive: true }` for `fs.mkdir` which has the same behavior. This necessitated bumping the supported node version to 10.12 as that's the version of Node that added this feature.

Note, I did not remove the recursive folder removal in the socket test (https://github.com/node-portfinder/node-portfinder/blob/b64ab4f707fccc079da3f86716ec0c78be61a30b/test/port-finder-socket.test.js#L68-L76) as `recursive` was added to [`fs.rmdir`](https://nodejs.org/api/fs.html#fsrmdirpath-options-callback) in 12.10, and then they deprecated that option and added the [`fs.rm`](https://nodejs.org/api/fs.html#fsrmpath-options-callback) function in 14.14.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,13 +7,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [6.x, 8.x, 10.x, 12.x, 14.x, 16.x, 18.x, 20.x, 22.x]
+        node-version: [10.x, 12.x, 14.x, 16.x, 18.x, 20.x, 22.x]
         os: [ubuntu-latest, macos-latest, windows-latest]
         exclude:
-          - os: macos-latest
-            node-version: 6.x
-          - os: macos-latest
-            node-version: 8.x
           - os: macos-latest
             node-version: 10.x
           - os: macos-latest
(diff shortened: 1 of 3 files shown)


eriktrom:
Awesome, thanks!

eriktrom:
> Note, I did not remove the recursive folder removal in the socket test (https://github.com/node-portfinder/node-portfinder/blob/b64ab4f707fccc079da3f86716ec0c78be61a30b/test/port-finder-socket.test.js#L68-L76) as recursive was added to [fs.rmdir](https://nodejs.org/api/fs.html#fsrmdirpath-options-callback) in 12.10, and then they deprecated that option and added the [fs.rm](https://nodejs.org/api/fs.html#fsrmpath-options-callback) function in 14.14.

Thanks for being so careful 👍

PR #168, opened 2025-03-12
getPort/getPorts return promise if no callback

### LINKED ISSUE #166: Combine callback and promise functions into one
Most libraries that I work with that support both callback and promises offer a unified interface where if you call the function with a callback it uses that, and if omit the callback then it returns a promise. The `async` package is an example of this (https://caolan.github.io/async/v3/docs.html#concat).  I think it makes for a cleaner API overall, vs having a separate function that does promises as currently exists for this library.

The type signature would look like:

```ts
export function getPort(callback: PortfinderCallback): void;
export function getPort(options: PortFinderOptions, callback: PortfinderCallback): void;
export function getPort(options?: PortFinderOptions): Promise<number>;
```

This would pose a breaking change for people who were calling the function without a callback, but given the nature of this library, I'm not sure why anyone would be doing that, and I think it's fine to break such a workflow. Would also get rid of the `*Promise` variants of functions, but the migration for people relying on them is pretty straightforward.

MasterOdin:
Closes #166 

PR makes it so that calling the `getPort` or `getPorts` method without providing a callback will have them return a promise instead. This makes the library more similar to others that support both a callback and a promise interface, without needing extra functions.

The `getPortPromise` and `getPortsPromise` methods are retained for backwards compatibility, but may wish to mark them as deprecated via jsdoc so that people move to using `getPort` and `getPorts`.

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -121,18 +121,12 @@ exports.basePath = '/tmp/portfinder'
 //
 exports.setBasePath = function (path) {
   exports.basePath = path;
-}
+};
 
-// ### function getPort (options, callback)
-// #### @options {Object} Settings to use when finding the necessary port
-// #### @callback {function} Continuation to respond to when complete.
-// Responds with a unbound port on the current machine.
-//
-exports.getPort = function (options, callback) {
+internals.getPort = function (options, callback) {
   if (!callback) {
     callback = options;
     options = {};
-
   }
 
   options.port   = Number(options.port) || Number(exports.basePort);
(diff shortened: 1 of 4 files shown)


test/port-finder.test.js
```diff
@@ -61,6 +61,20 @@ describe('with 5 existing servers', function () {
...
+  ])('the %s promise method should respond with the first free port (32773)', function (name, method, done) {
+    method()
+      .then(function (port) {
+        expect(port).toEqual(32773);
+        done();
+      })
+      .catch(function (err) {
+        done(err);
+      });
+  });
```
eriktrom, line 77: TY

lib/portfinder.js
```diff
@@ -220,33 +214,39 @@ exports.getPort = function (options, callback) {
...
     options = {};
   }
-  return new Promise(function(resolve, reject) {
-    exports.getPort(options, function(err, port) {
-      if (err) {
-        return reject(err);
-      }
-      resolve(port);
+
+  if (!callback) {
```
eriktrom, line 228: very good call btw

eriktrom:
One of my favorite PRs ever. So clean. One API. Thank you! 

> The getPortPromise and getPortsPromise methods are retained for backwards compatibility, but may wish to mark them as deprecated via jsdoc so that people move to using getPort and getPorts

Yeah - I mean - we just don't ever need to remove the aliases, given the code to keep them is a one liner and not subject to maintenance issues - and yeah the same idea applies to `getSocket` per #169 (see comment there)

eriktrom:
closes #166

PR #178, opened 2025-03-14
Update error message: output port at which port finder failed

eriktrom:
Update error message: output port at which port finder failed

First attempt to modernize error messages for both callback and promise, and make them more accurate on failure.

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -198,7 +198,7 @@ internals.getPort = function (options, callback) {
         return callback(null, openPorts[0]);
       }
       else {
-        const msg = 'No open ports found in between '+ options.startPort + ' and ' + options.stopPort;
+        const msg = `Searching for an open port failed at port ${openPorts[0]}.`;
         return callback(Error(msg));
       }
     } else {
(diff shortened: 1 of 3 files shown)


eriktrom:
review me @MasterOdin

lib/portfinder.js
```diff
@@ -198,7 +198,7 @@ internals.getPort = function (options, callback) {
         return callback(null, openPorts[0]);
       }
       else {
-        const msg = 'No open ports found in between '+ options.startPort + ' and ' + options.stopPort;
+        const msg = `Searching for an open port failed at port ${openPorts[0]}.`;
```
MasterOdin, line 201: I think this makes the error message worse for the single port case as it's less informative, where it sort of sounds like the library only tested port 32773 and that failed, so here's an error, where it had checked several ports.

For the multiple port case, I think it's fine for the callback variant as you get a partial array of ports it could find, whereas for the promise variant I think it's worse as similar to the single port variant, you cannot tell that it had tested other ports before failing.

I think ideally the error message would be as it is for single port and then like "Could not find 3 open ports between ...." as then it still gives the range info.

eriktrom: Yeah I agree. Whether we wanted to change the error message and how is basically what I wanted to get your feedback on. 

Thanks again for helping out. It’s a lot more likely to change and get better with two heads.

PR #179, opened 2025-03-15
Add promise return for getSocket

### LINKED ISSUE #169: Add promise variant of `getSocket` method
There's a promise variant of `getPort` and `getPorts`, there should then also be a promise variant of `getSocket` as well.

@eriktrom If this gets done, would you want to see both a `getSocketPromise` method as well as having `getSocket` be able to return a promise (assuming #168 is merged as that lays out the implementation pattern).

MasterOdin:
Closes #169 

Adds a promise return for `getSocket` as well as `getSocketPromise` function, so as to match the `getPort` and `getPorts` functions. Also adds type information for these functions as that was totally missing before.

The overall tests that are run remain the same, just that they're run against all three variants as was setup in #172.

### DIFF
--- a/lib/portfinder.js
+++ b/lib/portfinder.js
@@ -232,7 +232,7 @@ exports.getPort = function (options, callback) {
       });
     });
   } else {
-    return internals.getPort(options, callback);
+    internals.getPort(options, callback);
   }
 };
 
@@ -286,7 +286,7 @@ exports.getPorts = function (count, options, callback) {
       });
     });
   } else {
-    return internals.getPorts(count, options, callback);
+    internals.getPorts(count, options, callback);
   }
 };
 
@@ -298,19 +298,7 @@ exports.getPorts = function (count, options, callback) {
 //
 exports.getPortsPromise = exports.getPorts;
 
-//
-// ### function getSocket (options, callback)
-// #### @options {Object} Settings to use when finding the necessary port
-// #### @callback {function} Continuation to respond to when complete.
-// Responds with a unbound socket using the specified directory and base
-// name on the current machine.
-//
-exports.getSocket = function (options, callback) {
-  if (!callback) {
-    callback = options;
-    options = {};
-  }
-
+internals.getSocket = function (options, callback) {
   options.mod  = options.mod    || parseInt(755, 8);
   options.path = options.path   || exports.basePath + '.sock';
 
@@ -337,7 +325,7 @@ exports.getSocket = function (options, callback) {
         // next socket.
         //
         options.path = exports.nextSocket(options.path);
-        exports.getSocket(options, callback);
+        internals.getSocket(options, callback);
       }
     });
   }
@@ -385,6 +373,37 @@ exports.getSocket = function (options, callback) {
     ? testSocket()
     : checkAndTestSocket();
 };
+
+//
+// ### function getSocket (options, callback)
+// #### @options {Object} Settings to use when finding the necessary port
+// #### @callback {function} Continuation to respond to when complete.
+// Responds with a unbound socket using the specified directory and base
+// name on the current machine.
+//
+exports.getSocket = function (options, callback) {
+  if (typeof options === 'function') {
+    callback = options;
+    options = {};
+  }
+
+  options = options || {};
+
+  if (!callback) {
+    return new Promise(function (resolve, reject) {
+      internals.getSocket(options, function (err, socketPath) {
+        if (err) {
+          return reject(err);
+        }
+        resolve(socketPath);
+      });
+    });
+  } else {
+    internals.getSocket(options, callback);
+  }
+}
+
+exports.getSocketPromise = exports.getSocket;
 
 //
 // ### function nextPort (port)
--- a/test/port-finder-socket.test.js
+++ b/test/port-finder-socket.test.js
@@ -25,7 +25,8 @@ function createServers (callback) {
     function (next) {
       const server = net.createServer(function () { }),
             name = base === 0 ? 'test.sock' : 'test' + base + '.sock';
-      let sock = path.join(socketDir, name);
+      const socket = path.join(socketDir, name);
+      let sock = socket;
 
       // shamelessly stolen from foreverjs,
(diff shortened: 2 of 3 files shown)


lib/portfinder.js
```diff
@@ -286,7 +286,7 @@ exports.getPorts = function (count, options, callback) {
       });
     });
   } else {
-    return internals.getPorts(count, options, callback);
+    internals.getPorts(count, options, callback);
```
MasterOdin, line 289: This and the change to `internals.getPort` above are no-ops in terms of what they do (as the functions don't return anything), but I feel this is more "semantically correct" in that we don't expect this function to return anything if not a promise.

test/port-finder-socket.test.js
```diff
@@ -88,60 +86,126 @@ describe('portfinder', function () {
...
-        portfinder.getSocket({
-          path: path.join(socketDir, 'exists.sock'),
-        }, function (err, socket) {
-          expect(err).toBeNull();
-          expect(socket).toEqual(path.join(socketDir, 'exists.sock'));
-          done();
-        });
+      // We don't use `test.sock` here due to some race condition on Windows in freeing the `test.sock` file
+      // when we close the servers.
+      test('with a directory that exists should respond with the first free socket (foo.sock)', function (done) {
```
MasterOdin, line 188: The test was mentioning `test.sock`, but the code was using `exists.sock`. I'm guessing this was due to an issue on windows where the socket files were not being cleared as part of `cleanup`. While this was fixed, I think there's still some weird race condition where calling this function too quickly will still fail on Windows if using `test.sock`.

As such, I went with calling it `foo.sock` to avoid the name collision and that I renamed it as I didn't like calling it `exists.sock` as that felt like it was implying the socket file already existed.

lib/portfinder.js
eriktrom: yeah makes sense

test/port-finder-socket.test.js
```diff
@@ -25,7 +25,8 @@ function createServers (callback) {
     function (next) {
       const server = net.createServer(function () { }),
             name = base === 0 ? 'test.sock' : 'test' + base + '.sock';
-      let sock = path.join(socketDir, name);
+      const socket = path.join(socketDir, name);
+      let sock = socket;
```
eriktrom, line 29: Just so I understand - why are we changing this and what is going on - e.g., why are we setting another variable for the socket handle

test/port-finder-socket.test.js
```diff
@@ -46,22 +47,19 @@ function createServers (callback) {
 
       server.listen(sock, next);
       base++;
-      servers.push(server);
+      servers.push([server, socket]);
```
eriktrom, line 50: also - why are we adding an array with `[server, socket`] here - vs what we did in the past?

test/port-finder-socket.test.js
```diff
@@ -46,22 +47,19 @@ function createServers (callback) {
...
-      if (err) {
-        callback(err, false);
-      } else {
-        stopServers(callback, index + 1);
+function stopServers(callback) {
+  _async.each(servers, function ([server, socket], next) {
+    server.close(function () {
+      if (process.platform === 'win32' && fs.existsSync(socket)) {
+        fs.unlinkSync(socket);
       }
```
eriktrom, line 59: oh I see - calling server.close doesn't work on windows? was this always something we just left dangling before? 

```js
      if (process.platform === 'win32') {
        //
        // Create 'symbolic' file on the system, so it can be later
        // found via "forever list" since the `\\.pipe\\*` "files" can't
        // be enumerated because ... Windows.
        //
        fs.openSync(sock, 'w');
```


Maybe just explain this all to me like a 5 years old - thanks

test/port-finder-socket.test.js
```diff
@@ -88,60 +85,124 @@ describe('portfinder', function () {
 
   describe('with 5 existing servers', function () {
     beforeAll(function (done) {
-      createServers(function () {
-        portfinder.getSocket({
-          path: path.join(badDir, 'test.sock'),
-        }, function () {
-          done();
-        });
```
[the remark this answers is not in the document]
eriktrom: that's very clean :)

test/port-finder-socket.test.js
eriktrom: to run `fs.unlinkSync(socket);` on windows, is that correct? (and we just never did that prior to this?)

test/port-finder-socket.test.js
eriktrom: again, to run `fs.unlinkSync(socket);` on Windows, which we didn't do before, it seems. If different let me know thanks

test/port-finder-socket.test.js
eriktrom: I guess I only have 1 q then - why to variables `socket` and `sock` for the same os handle?

test/port-finder-socket.test.js
eriktrom: okay cool - might answer why were now calling `fs.unlinkSync(socket);` as well it seems

thanks for the note

test/port-finder-socket.test.js
eriktrom: related does calling `fs.openSync(sock, 'w');` with `socket` instead of `sock` work? (below, not part of diff)

eriktrom:
I cleaned up my comments. I see windows is a little feisty :) 

Just one question about variable naming. The rest all makes sense. 

This is awesome work, thank you very much, I really appreciate it 👍

test/port-finder-socket.test.js
MasterOdin: If the platform is windows, we append the pipe stuff to the front of `sock` to be passed to `server.listen`. However, we create the file without the pipe stuff, so that's why I've used the two separate variables, `socket` to track the filename as generated, and then `sock` to be modified with the pipe stuff.

eriktrom: sound good

eriktrom:
Thanks again @MasterOdin

